PR #2, opened 2026-08-12
Add OpenRouter BYOK support and harden TodoWrite

0vp:
## Summary
- add OpenRouter as a fully supported BYOK provider, including authenticated model discovery, key validation, streaming tool calls, usage reporting, thinking controls, and TUI key management
- preserve and replay opaque OpenRouter reasoning metadata across tool continuations and durable session resume
- make TodoWrite normalize common model-generated states instead of rejecting them, prevent same-turn false completion, and safely restore only successfully executed todo updates

## Verification
- exercised live SSO tool continuations with OpenRouter, Google/Gemini, Anthropic, and OpenAI
- exercised live BYOK tool continuations with OpenRouter, Google/Gemini, Anthropic, and OpenAI
- verified BYOK and SSO compaction plus post-compaction continuation
- reproduced the reported OpenAI TodoWrite flow in the TUI and confirmed unfinished work remains active
- `bun run validate`: 1,352 passed, 3 skipped, 0 failed
- `bun run build`

### DIFF
--- a/src/core/agent/SubAgentRunner.ts
+++ b/src/core/agent/SubAgentRunner.ts
@@ -34,6 +34,7 @@ export class SubAgentRunner {
 		const bus = new EventBus();
 		const session = new Session(shortId("subagent"));
 		const detachSessionProjection = session.attach(bus);
+		const turnStartTodos = [...session.todos];
 		const childLog = params.trace
 			? new ClientEventLog(params.trace.sessionId, params.trace.clientLogPath)
 			: null;
@@ -98,6 +99,7 @@ export class SubAgentRunner {
 				throw new Error("A sub-agent cannot ask the user a question.");
 			},
 			getTodos: () => session.todos,
+			getTurnStartTodos: () => turnStartTodos,
 			agentDepth: params.depth,
 			trace: params.trace?.context,
 			lsp: this.deps.lsp,
--- a/src/core/agent/SubAgentRunner.ts
+++ b/src/core/agent/SubAgentRunner.ts
@@ -34,7 +34,6 @@ export class SubAgentRunner {
 		const bus = new EventBus();
 		const session = new Session(shortId("subagent"));
 		const detachSessionProjection = session.attach(bus);
-		const turnStartTodos = [...session.todos];
 		const childLog = params.trace
 			? new ClientEventLog(params.trace.sessionId, params.trace.clientLogPath)
 			: null;
@@ -99,7 +98,6 @@ export class SubAgentRunner {
 				throw new Error("A sub-agent cannot ask the user a question.");
 			},
 			getTodos: () => session.todos,
-			getTurnStartTodos: () => turnStartTodos,
 			agentDepth: params.depth,
 			trace: params.trace?.context,
 			lsp: this.deps.lsp,
--- a/src/core/todos/TodoList.ts
+++ b/src/core/todos/TodoList.ts
@@ -4,6 +4,50 @@ import type { Message } from "../session/Message.ts";
 import { canonicalToolName } from "../tools/names.ts";
 import type { TodoDraft } from "./types.ts";
 
+/**
+ * Models occasionally emit a structurally valid list with no active item, or
+ * more than one. Keep the provider contract forgiving while making the stored
+ * UI state deterministic: pending work always has exactly one active item.
+ */
+export function normalizeTodoDrafts(drafts: readonly TodoDraft[]): TodoDraft[] {
+	const normalized = drafts.map((draft) => ({ ...draft }));
+	const activeIndexes = normalized.flatMap((todo, index) =>
+		todo.status === "in_progress" ? [index] : [],
+	);
+
+	if (activeIndexes.length > 1) {
+		for (const index of activeIndexes.slice(1)) {
+			const todo = normalized[index];
+			if (todo) todo.status = "pending";
+		}
+	}
+
+	const hasActive = normalized.some((todo) => todo.status === "in_progress");
+	if (!hasActive) {
+		const firstPending = normalized.find((todo) => todo.status === "pending");
+		if (firstPending) firstPending.status = "in_progress";
+	}
+	return normalized;
+}
+
+/**
+ * Preserve completion transitions for existing todos, but do not let a model
+ * introduce brand-new work as already complete.
+ */
+export function normalizeTodoUpdate(
+	drafts: readonly TodoDraft[],
+	previousTodos: readonly TodoItem[],
+): TodoDraft[] {
+	const previousContents = new Set(previousTodos.map((todo) => todo.content));
+	return normalizeTodoDrafts(
+		drafts.map((draft) =>
+			draft.status === "completed" && !previousContents.has(draft.content)
+				? { ...draft, status: "pending" }
+				: draft,
+		),
+	);
+}
+
 export function reconcileTodos(
 	drafts: readonly TodoDraft[],
 	previousTodos: readonly TodoItem[],
@@ -39,11 +83,29 @@ export function todosFromMessages(messages: readonly Message[]): TodoItem[] {
 		for (let j = message.toolCalls.length - 1; j >= 0; j--) {
 			const call = message.toolCalls[j];
 			if (!call || canonicalToolName(call.name) !== "todo_write") continue;
+			const matchingResult = findToolResult(messages, i + 1, call.id);
+			if (!matchingResult || matchingResult.isError) continue;
 			const todos = todoItemsFromInput(call.input);
 			return areTodosComplete(todos) ? [] : todos;
 		}
 	}
 	return [];
+}
+
+function findToolResult(
+	messages: readonly Message[],
+	startIndex: number,
+	toolCallId: string,
+) {
+	for (let i = startIndex; i < messages.length; i++) {
+		const message = messages[i];
+		if (message?.role !== "tool") break;
+		const result = message.results.find(
+			(candidate) => candidate.toolCallId === toolCallId,
+		);
+		if (result) return result;
+	}
+	return undefined;
 }
 
 function todoItemsFromInput(input: unknown): TodoItem[] {
@@ -66,5 +128,8 @@ function todoItemsFromInput(input: unknown): TodoItem[] {
 			continue;
 		todos.push({ id: shortId("todo"), content, status });
 	}
-	return todos;
+	return normalizeTodoDrafts(todos).map((todo) => ({
+		id: shortId("todo"),
+		...todo,
+	}));
 }
--- a/src/core/todos/TodoList.ts
+++ b/src/core/todos/TodoList.ts
@@ -9,7 +9,9 @@ import type { TodoDraft } from "./types.ts";
  * more than one. Keep the provider contract forgiving while making the stored
  * UI state deterministic: pending work always has exactly one active item.
  */
-export function normalizeTodoDrafts(drafts: readonly TodoDraft[]): TodoDraft[] {
+export function normalizeTodoDrafts<T extends TodoDraft>(
+	drafts: readonly T[],
+): T[] {
 	const normalized = drafts.map((draft) => ({ ...draft }));
 	const activeIndexes = normalized.flatMap((todo, index) =>
 		todo.status === "in_progress" ? [index] : [],
@@ -128,8 +130,5 @@ function todoItemsFromInput(input: unknown): TodoItem[] {
 			continue;
 		todos.push({ id: shortId("todo"), content, status });
 	}
-	return normalizeTodoDrafts(todos).map((todo) => ({
-		id: shortId("todo"),
-		...todo,
-	}));
+	return normalizeTodoDrafts(todos);
 }
--- a/src/providers/byok/ByokClient.ts
+++ b/src/providers/byok/ByokClient.ts
@@ -327,6 +327,7 @@ export class ByokClient implements AgentClient {
 		let terminal = false;
 		let events = 0;
 		let error: string | undefined;
+		let providerMetadata: string | undefined;
 
 		const commit = async (): Promise<string | null> => {
 			if (committed) return null;
@@ -336,6 +337,7 @@ export class ByokClient implements AgentClient {
 					role: "assistant",
 					content: text,
 					toolCalls: calls,
+					...(providerMetadata ? { providerMetadata } : {}),
 				});
 			}
 			const threadId = options.cacheKey;
@@ -378,6 +380,7 @@ export class ByokClient implements AgentClient {
 				}
 				if (event.kind === "requires_action") {
 					terminal = true;
+					providerMetadata = event.providerMetadata;
 					const warning = await commit();
 					if (warning) yield { kind: "warning", message: warning };
 				}
@@ -473,11 +476,7 @@ export class ByokClient implements AgentClient {
 		};
 	}
 
-	/**
-	 * Answered locally. Backboard exposes per-model thinking metadata over HTTP;
-	 * vendors do not, so this reports what the adapter's catalog already knows
-	 * and lets the thinking resolver fall back to its defaults.
-	 */
+	/** Resolve provider-specific thinking support with the active BYOK key. */
 	async getModelThinkingMetadata(
 		provider: string,
 		model: string,
@@ -486,7 +485,9 @@ export class ByokClient implements AgentClient {
 		return {
 			provider,
 			model,
-			supports_thinking: adapter?.supportsThinking(model) ?? false,
+			supports_thinking: adapter
+				? await adapter.supportsThinking(model, this.keyFor(adapter))
+				: false,
 		};
 	}
 
--- a/src/providers/byok/adapters/OpenRouterAdapter.ts
+++ b/src/providers/byok/adapters/OpenRouterAdapter.ts
@@ -1,0 +1,525 @@
+import { createHash } from "node:crypto";
+import type { ThinkingConfig } from "../../../config/defaults.ts";
+import type {
+	ModelCatalogItem,
+	ProviderEvent,
+	ProviderToolCall,
+} from "../../backboard/types.ts";
+import { unexpectedStreamEndMessage } from "../ByokError.ts";
+import type {
+	ByokMessage,
+	ByokStreamRequest,
+	ProviderAdapter,
+} from "../ByokTypes.ts";
+import { getJson, postSseJson } from "../httpStream.ts";
+import {
+	OPENROUTER_API_BASE,
+	OPENROUTER_APP_TITLE,
+	OPENROUTER_APP_URL,
+	OPENROUTER_CATALOG_TTL_MS,
+	OPENROUTER_REASONING_PARAMETERS,
+	OPENROUTER_REQUIRED_MODEL_PARAMETERS,
+} from "./OpenRouterAdapter.constants.ts";
+
+interface OpenRouterModelsResponse {
+	data?: OpenRouterModel[];
+}
+
+interface OpenRouterModel {
+	id?: string;
+	created?: number;
+	context_length?: number;
+	top_provider?: { max_completion_tokens?: number | null };
+	supported_parameters?: string[];
+	architecture?: {
+		input_modalities?: string[];
+		output_modalities?: string[];
+	};
+}
+
+interface PendingToolCall {
+	id: string;
+	name: string;
+	args: string;
+	announced: boolean;
+}
+
+interface OpenRouterCatalog {
+	models: ModelCatalogItem[];
+}
+
+interface CachedOpenRouterCatalog {
+	credentialId: string;
+	expiresAt: number;
+	value?: OpenRouterCatalog;
+	pending?: Promise<OpenRouterCatalog>;
+}
+
+let cachedCatalog: CachedOpenRouterCatalog | null = null;
+
+export const openRouterAdapter: ProviderAdapter = {
+	id: "openrouter",
+	label: "OpenRouter",
+	consoleUrl: "https://openrouter.ai/settings/keys",
+	keyHint: "sk-or-v1-...",
+
+	looksLikeKey(key) {
+		return /^sk-or-v1-[\w-]{20,}$/.test(key.trim());
+	},
+
+	async validateKey(key, signal) {
+		await getJson<unknown>(
+			`${OPENROUTER_API_BASE}/key`,
+			headers(key),
+			"openrouter",
+			signal,
+		);
+	},
+
+	async listModels(key, signal) {
+		return (await loadCatalog(key, signal)).models;
+	},
+
+	async supportsThinking(model, key) {
+		const catalog = await loadCatalog(key);
+		return (
+			catalog.models.find((candidate) => candidate.name === model)
+				?.supports_thinking === true
+		);
+	},
+
+	stream(request, key) {
+		return streamOpenRouter(request, key);
+	},
+};
+
+export function isOpenRouterChatModel(model: OpenRouterModel): boolean {
+	if (!model.id) return false;
+	const parameters = new Set(model.supported_parameters ?? []);
+	if (
+		!OPENROUTER_REQUIRED_MODEL_PARAMETERS.every((parameter) =>
+			parameters.has(parameter),
+		)
+	) {
+		return false;
+	}
+	const input = model.architecture?.input_modalities;
+	const output = model.architecture?.output_modalities;
+	return (
+		(input?.includes("text") ?? true) && (output?.includes("text") ?? true)
+	);
+}
+
+export function supportsOpenRouterThinking(
+	supportedParameters: readonly string[] | undefined,
+): boolean {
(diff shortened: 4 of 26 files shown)


src/tools/TodoWriteTool.tsx
```diff
@@ -96,9 +72,11 @@ export class TodoWriteTool extends Tool<Input, Output> {
 		input: Input,
 		ctx: ToolContext,
 	): Promise<ToolResult<Output>> {
+		const previousTodos = ctx.getTodos?.() ?? [];
+		const completableTodos = ctx.getTurnStartTodos?.() ?? previousTodos;
```
muhammadbalawal: **[1 — High] Completion is gated on the turn-start snapshot, so a todo created during the current turn can never be completed in that turn.**

`AgentController.runTurn` captures `turnStartTodos = [...session.todos]` (line 299) *before* the model has written any plan, so on a fresh session it is `[]`.

Concrete sequence:
1. Model calls TodoWrite `[A, B, C]` all pending.
2. Model does the work for A.
3. Model calls TodoWrite `[A completed, B in_progress, C pending]`.
4. `normalizeTodoUpdate` sees A's content is absent from the *empty* turn-start set and rewrites A back to `pending`.

The list never advances. Worse, the turn then ends with non-completed todos, so `todoReconciliationNotification` fires (`shouldFire: todos.some(status !== "completed")`) and re-prompts the model with the full tool set — every corrective TodoWrite is reverted the same way, burning all `TODO_RECONCILIATION_MAX_PASSES = 5` extra model round-trips on *every* turn.

The new test `"blocks completion of a todo introduced earlier in the same turn"` locks this behaviour in, so it reads as intentional — but it breaks the single-turn plan-and-execute flow that TodoWrite exists for.

**Suggested fix:** pass `previousTodos` (the live `ctx.getTodos()`) as the completable set. That still blocks the actual abuse case — a model introducing brand-new work already marked `completed` in a single call — because such an item is absent from the live list too.

src/core/agent/SubAgentRunner.ts
```diff
@@ -34,6 +34,7 @@ export class SubAgentRunner {
 		const bus = new EventBus();
 		const session = new Session(shortId("subagent"));
 		const detachSessionProjection = session.attach(bus);
+		const turnStartTodos = [...session.todos];
```
muhammadbalawal: **[2 — High] Sub-agents can never complete any todo, ever.**

`session` is constructed one line above:

```ts
const session = new Session(shortId("subagent"));
const detachSessionProjection = session.attach(bus);
const turnStartTodos = [...session.todos];   // always []
```

So `turnStartTodos` is permanently `[]`, and unlike the main agent there is no later user turn to refresh it. Every `completed` draft a sub-agent writes is rewritten to `pending` by `normalizeTodoUpdate`, `areTodosComplete` never returns true, and the sub-agent's todo panel shows work that never finishes.

Even under the intent behind finding #1, this case is unambiguously wrong — there is no "next turn" to rescue it.

src/core/todos/TodoList.ts
```diff
@@ -4,6 +4,50 @@ import type { Message } from "../session/Message.ts";
...
+
+/**
+ * Preserve completion transitions for existing todos, but do not let a model
+ * introduce brand-new work as already complete.
+ */
+export function normalizeTodoUpdate(
+	drafts: readonly TodoDraft[],
+	previousTodos: readonly TodoItem[],
+): TodoDraft[] {
+	const previousContents = new Set(previousTodos.map((todo) => todo.content));
```
muhammadbalawal, line 43: **[3 — High] Todo identity is the exact `content` string, so any rewording silently reverts a completion.**

```ts
const previousContents = new Set(previousTodos.map((todo) => todo.content));
```

Turn 1 stores `{content: "Add tests", status: in_progress}`. Turn 2 the model submits `{content: "Add tests for the parser", status: "completed"}` — a routine paraphrase. No exact match in `previousContents`, so the item is forced to `pending`, then `reconcileTodos` (also content-keyed) mints a fresh id.

The user watches a finished item bounce back to pending with no explanation, and the reconciliation reminder fires again. Content-string equality is too brittle to carry state-transition authority.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+	const body: Record<string, unknown> = {
+		model: request.model,
+		messages: toOpenRouterMessages(request),
+		stream: true,
+		stream_options: { include_usage: true },
+	};
+	if (request.tools.length > 0) {
+		body.tools = request.tools;
+		body.tool_choice = "auto";
+		body.provider = { require_parameters: true };
```
muhammadbalawal, line 140: **[4 — Medium] `require_parameters: true` combined with an unconditionally-attached `reasoning` field turns a stale/failed catalog into a hard routing failure.**

`isOpenRouterChatModel` only requires `tools`, so non-reasoning models are in the catalog. Meanwhile `createRuntimeThinkingResolver` wraps the metadata fetch in `.catch(() => null)`, and `resolveThinkingCapabilities` treats null metadata as supported (`metadata?.supports_thinking !== false`).

So when `/models/user` is down, rate-limited, or the model is newly added, the request carries `reasoning: {max_tokens: N}` **and** `provider: {require_parameters: true}`. OpenRouter then refuses with *"No endpoints found that support your parameters"* rather than routing to a provider that would have served the turn fine without the reasoning field.

Consider only setting `require_parameters` when the catalog actually confirmed the model's parameter set, or dropping `reasoning` when `supports_thinking` is unknown.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+			Array.isArray(detail)
+		) {
+			continue;
+		}
+		const incoming = detail as Record<string, unknown>;
+		const type = typeof incoming.type === "string" ? incoming.type : "unknown";
+		const index =
+			typeof incoming.index === "number" && Number.isFinite(incoming.index)
+				? incoming.index
+				: pending.size;
```
muhammadbalawal: **[5 — Medium] `pending.size` as the index fallback fragments streamed reasoning into one entry per chunk.**

```ts
const index =
  typeof incoming.index === "number" && Number.isFinite(incoming.index)
    ? incoming.index
    : pending.size;
```

When a provider streams `reasoning_details` without an `index` field (only `type`/`text`):

- Chunk 1: map empty → key `0:reasoning.text`, insert → size becomes 1
- Chunk 2: size is 1 → key `1:reasoning.text` → **new** entry
- Chunk 3: key `2:reasoning.text` → **new** entry

The text-concatenation branch below never runs, and the `providerMetadata` replayed on the next tool continuation is N fragmented reasoning blocks instead of one. For signature-verified reasoning (Anthropic via OpenRouter) the replay is rejected or the reasoning is discarded.

A stable fallback (e.g. a per-`type` counter fixed at first sight, or just `type` alone when `index` is absent) avoids the drift.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+	key: string,
+	signal?: AbortSignal,
+): Promise<OpenRouterCatalog> {
+	const normalizedKey = key.trim();
+	const credentialId = createHash("sha256").update(normalizedKey).digest("hex");
+	if (cachedCatalog?.credentialId === credentialId) {
+		if (cachedCatalog.value && cachedCatalog.expiresAt > Date.now()) {
+			return cachedCatalog.value;
+		}
+		if (cachedCatalog.pending) return cachedCatalog.pending;
```
muhammadbalawal, line 306: **[6 — Medium] The shared in-flight promise cross-cancels unrelated callers, and the module-global leaks across clients and tests.**

The `/model` picker calls `listModels(key, signal)`, which stores `pending = fetchCatalog(key, signal)` in `cachedCatalog`. Concurrently the turn's `getModelThinkingMetadata` calls `supportsThinking(model, key)` → `loadCatalog(key)` **with no signal**, and is handed the same promise on this line.

User presses Esc → picker aborts → the thinking-metadata lookup, which never opted into cancellation, rejects with `AbortError`.

Two further consequences of `cachedCatalog` being a module-level global:
- State is shared across all `ByokClient` instances, not scoped to one.
- It leaks across tests in `OpenRouterAdapter.test.ts` — the same stub key string produces the same sha256, so the 5-minute TTL makes assertions like `expect(captured).toHaveLength(1)` order-dependent. There is no exported reset hook.

At minimum, don't propagate one caller's `signal` into a promise other callers share.

muhammadbalawal (commented):
## Code review — 14 findings

Automated recall-oriented review of `origin/main...qli/rcli/openrouter` (24 files, +1482/-77). Inline comments below, ranked most severe first: **3 blocking correctness bugs in the TodoWrite hardening**, **6 correctness/robustness issues in the OpenRouter adapter**, and **5 cleanup / dead-code items**.

### Headline

The TodoWrite change gates completion on a *turn-start* snapshot of the todo list. Because the snapshot is taken before the model has written any plan, **a todo created and finished inside the same turn can never be marked completed** — and for sub-agents the snapshot is permanently empty, so they can never complete a todo at all. This also drives `todoReconciliationNotification` into its full 5-pass retry loop on every turn, since the list can never reach an all-completed state.

| # | Severity | Location | Issue |
|---|---|---|---|
| 1 | High | `TodoWriteTool.tsx:76` | Turn-start snapshot blocks all same-turn completions |
| 2 | High | `SubAgentRunner.ts:37` | Sub-agent snapshot is always empty |
| 3 | High | `TodoList.ts:41` | Todo identity keyed on exact content string |
| 4 | Medium | `OpenRouterAdapter.ts:132` | `require_parameters` + unconditional `reasoning` |
| 5 | Medium | `OpenRouterAdapter.ts:487` | `pending.size` index fallback fragments reasoning |
| 6 | Medium | `OpenRouterAdapter.ts:299` | Shared in-flight promise / module-global cache |
| 7 | Medium | `OpenRouterAdapter.ts:372` | `{effort: undefined}` for empty ThinkingConfig |
| 8 | Medium | `OpenRouterAdapter.ts:276` | Wrong attribution header |
| 9 | Medium | `OpenRouterAdapter.ts:345` | OpenRouter models stuck at 128k context |
| 10 | Medium | `ByokClient.ts:489` | Eager `keyFor()` throws for key-less adapters |
| 11 | Low | `TodoList.ts:132` | Dead `id: shortId("todo")` |
| 12 | Low | `TodoList.ts:86` | Todos dropped when result record is missing |
| 13 | Low | `OpenRouterAdapter.ts:136` | `maxOutputTokens` branch unreachable |
| 14 | Low | `OpenRouterAdapter.ts:58` | Duplicates `utils/asyncCache.ts` |

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+): Record<string, unknown> | null {
+	if (!thinking) return null;
+	if ("max_tokens" in thinking) {
+		return { max_tokens: thinking.max_tokens };
+	}
+	if ("budget_tokens" in thinking) {
+		return { max_tokens: thinking.budget_tokens };
+	}
+	return {
+		effort: thinking.effort === "max" ? "high" : thinking.effort,
```
muhammadbalawal, line 382: **[7 — Medium] Empty `ThinkingConfig` produces `{ effort: undefined }` → serializes to `"reasoning": {}`.**

`ThinkingConfig` includes `Record<string, never>`, and `resolveThinking` returns exactly `{}` for defaults-only models and for deferred dynamic thinking.

That value is truthy, has neither `max_tokens` nor `budget_tokens`, so it falls through to this branch and reads `thinking.effort` = `undefined`. Then `if (reasoning) body.reasoning = reasoning` attaches it, and `JSON.stringify` drops the undefined key — sending `reasoning: {}`.

Result: reasoning switched **on** where the resolver meant "leave provider defaults alone", plus an extra parameter that `require_parameters: true` (see #4) now demands the routed endpoint support.

An explicit `if (!("effort" in thinking)) return null;` guard before this return closes it.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+		return;
+	}
+	yield { kind: "completed", usage: finalUsage };
+}
+
+function headers(key: string): Record<string, string> {
+	return {
+		Authorization: `Bearer ${key.trim()}`,
+		"HTTP-Referer": OPENROUTER_APP_URL,
+		"X-OpenRouter-Title": OPENROUTER_APP_TITLE,
```
muhammadbalawal, line 286: **[8 — Medium] Wrong attribution header — OpenRouter's documented header is `X-Title`, not `X-OpenRouter-Title`.**

`HTTP-Referer: https://backboard.io` is correct, but `X-OpenRouter-Title` is not a header OpenRouter recognises, so it is ignored and R-CLI never appears by name in the user's OpenRouter activity/leaderboard attribution.

Note the new test asserts the wrong header verbatim:

```ts
expect(captured[0]?.headers["X-OpenRouter-Title"]).toBe("Backboard R-CLI");
```

so it will not catch this. Both the header and the assertion need to change to `X-Title`.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+		const supportsThinking = supportsReasoningParameter(
+			model.supported_parameters,
+		);
+		models.push({
+			name: id,
+			provider: "openrouter",
+			model_type: "llm",
+			last_updated:
+				typeof model.created === "number" ? model.created * 1000 : null,
+			context_limit: positiveNumber(model.context_length),
```
muhammadbalawal, line 354: **[9 — Medium] The real `context_length` fetched here never reaches the running session — every OpenRouter model is treated as a 128k window.**

`ContextReport` calls `resolveContextWindow(model, session.reportedContextLimit)`. BYOK adapters emit no `contextUsage`, and `contextLimit` only reaches the session through the `/model` picker (`src/ui/App.tsx:956`).

So on startup with the new default `openrouter/anthropic/claude-sonnet-4.6` — or any `--model openrouter/...` — `contextWindowFor` lowercases to `anthropic/claude-sonnet-4.6`, which matches **no** `WINDOW_PATTERNS` entry (`^claude-` is anchored, and every OpenRouter id carries a `vendor/` prefix). It falls through to `DEFAULT_CONTEXT_WINDOW = 128_000`.

A 200k/1M-context model then auto-compacts at `AUTO_COMPACT_THRESHOLD_PERCENT` of 128k, discarding context that never needed to be dropped. Unlike the other BYOK adapters — whose bare model ids do match the table — OpenRouter can never hit it.

src/providers/byok/ByokClient.ts
```diff
@@ -486,7 +485,9 @@ export class ByokClient implements AgentClient {
 		return {
 			provider,
 			model,
-			supports_thinking: adapter?.supportsThinking(model) ?? false,
+			supports_thinking: adapter
+				? await adapter.supportsThinking(model, this.keyFor(adapter))
```
muhammadbalawal, line 489: **[10 — Medium] Eager `keyFor()` throws for the three adapters that don't need a key, inverting a previously reliable local answer.**

`this.keyFor(adapter)` is evaluated before the call, and it throws when no enabled key exists for that provider.

In key-only mode `ClientRouter.sourceFor` falls back to `"byok"` when no Backboard client exists, *even with no key for that provider*. So a user holding only an OpenRouter key who selects `google/gemma-4-31b-it`:

- **Before:** `supportsGoogleThinking(model)` — a pure function — correctly returned `supports_thinking: false`.
- **After:** `keyFor` throws → `createRuntimeThinkingResolver` swallows it with `.catch(() => null)` → `resolveThinkingCapabilities` treats null metadata as supported (`metadata?.supports_thinking !== false`).

That is the exact inversion of the answer the adapter could still have given for free. Passing a lazy resolver (`() => string`) instead of an eager key, or letting adapters declare whether they need one, keeps the static answers working.

src/core/todos/TodoList.ts
```diff
@@ -66,5 +128,8 @@ function todoItemsFromInput(input: unknown): TodoItem[] {
 			continue;
 		todos.push({ id: shortId("todo"), content, status });
 	}
-	return todos;
+	return normalizeTodoDrafts(todos).map((todo) => ({
+		id: shortId("todo"),
```
muhammadbalawal: **[11 — Low] Dead code: `id: shortId("todo")` is always overwritten by the spread.**

```ts
return normalizeTodoDrafts(todos).map((todo) => ({
  id: shortId("todo"),
  ...todo,
}));
```

Line 129 already pushes `{id: shortId("todo"), content, status}`, and `normalizeTodoDrafts` shallow-copies each entry (preserving `id`). Since `id:` is written *before* the spread, `todo.id` always wins.

Every hydrated todo pays a wasted `shortId()` call, and the line reads as if it were re-keying items when it provably cannot. Either drop the `id:` key or move the spread first if re-keying was the intent.

src/core/todos/TodoList.ts
```diff
@@ -39,13 +83,31 @@ export function todosFromMessages(messages: readonly Message[]): TodoItem[] {
 		for (let j = message.toolCalls.length - 1; j >= 0; j--) {
 			const call = message.toolCalls[j];
 			if (!call || canonicalToolName(call.name) !== "todo_write") continue;
+			const matchingResult = findToolResult(messages, i + 1, call.id);
```
muhammadbalawal, line 88: **[12 — Low] Requiring an adjacent non-error result also drops todos that were genuinely applied.**

`findToolResult` scans forward from `i + 1` and `break`s at the first non-`tool` message. If the process is killed — or the client log truncated — between writing the assistant message and its tool-results message, the resumed session finds no matching result and returns `[]`.

Todos the user watched execute, and that were live in session state before the crash, vanish on resume where the previous code restored them. The same applies to any transcript rewrite that separates a `todo_write` call from its result message.

The guard is right for the "never executed" case it targets; it just can't distinguish that from "result record lost".

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+		stream_options: { include_usage: true },
+	};
+	if (request.tools.length > 0) {
+		body.tools = request.tools;
+		body.tool_choice = "auto";
+		body.provider = { require_parameters: true };
+	}
+	const reasoning = openRouterReasoning(request.thinking);
+	if (reasoning) body.reasoning = reasoning;
+	if (request.maxOutputTokens) body.max_tokens = request.maxOutputTokens;
```
muhammadbalawal, line 144: **[13 — Low] `request.maxOutputTokens` is never populated — this branch is unreachable.**

`ByokClient.streamTurn` builds the adapter request from `model` / `systemPrompt` / `tools` / `messages` / `thinking` / `cacheKey` / `signal` only. `maxOutputTokens` is never set on any code path, so `body.max_tokens` is never sent.

Unlike `AnthropicAdapter` — which supplies `DEFAULT_MAX_OUTPUT_TOKENS` as a fallback — OpenRouter gets no cap at all, and this line gives a false impression that the CLI bounds output length. Either wire `maxOutputTokens` through `streamTurn` or drop the branch.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -0,0 +1,525 @@
...
+}
+
+interface CachedOpenRouterCatalog {
+	credentialId: string;
+	expiresAt: number;
+	value?: OpenRouterCatalog;
+	pending?: Promise<OpenRouterCatalog>;
+}
+
+let cachedCatalog: CachedOpenRouterCatalog | null = null;
```
muhammadbalawal, line 58: **[14 — Low] Re-implements `src/utils/asyncCache.ts` as a mutable module global.**

`createAsyncCache` already provides exactly the empty/pending/ready state machine and the "clear the slot when the promise rejects" behaviour that `loadCatalog` re-derives across lines 289-320 — including the fragile identity check:

```ts
if (cachedCatalog?.pending === pending) cachedCatalog = null;
```

Wrapping `createAsyncCache` in a small keyed + TTL holder owned by the adapter object would remove the duplicated state machine, remove the cross-instance global (see #6), and give tests the `clear()` hook they currently lack.

src/providers/byok/adapters/OpenRouterAdapter.ts
muhammadbalawal: **Retracted — this finding was wrong. The code is correct as written; please disregard.**

I checked the current OpenRouter API reference and `X-OpenRouter-Title` is the documented primary header:

> `HTTP-Referer`: Identifies your app on openrouter.ai
> `X-OpenRouter-Title`: Sets/modifies your app's title (`X-Title` also accepted)
> `X-OpenRouter-Categories`: Assigns marketplace categories

`X-Title` is the accepted alias, not the canonical name — I had the older docs in mind, where it was primary. Both headers work, and this PR uses the one the current reference leads with. No change needed here, and the test assertion is correct too.

Apologies for the noise. This was the one finding in the review with no repo evidence behind it.

muhammadbalawal:
## Follow-up: docs verified against `3557983` — one retraction, two new findings

I went back to the OpenRouter docs to check the two findings in my review that rested on recalled API knowledge rather than repo evidence. Results below.

### Verification of `3557983`

`bunx tsc --noEmit` clean · `biome check .` clean (606 files) · `bun test` **1356 pass / 3 skip / 0 fail**. This repo has no CI workflow, so that run is the only gate.

Findings **#1, #2, #5, #6, #7, #9, #11** are fixed. Removing the `getTurnStartTodos` plumbing outright rather than patching around it was the right call, and the `#6` fix additionally closes a stale-fetch-clobbers-newer-key race I had not flagged.

---

### ❌ #8 retracted — I was wrong

`X-OpenRouter-Title` is the **documented primary header**; `X-Title` is the accepted alias. The code was correct. Retracted [on the thread](https://github.com/Backboard-io/Backboard-R-CLI/pull/2#discussion_r3771602861) — no change needed.

### ✅ #4 confirmed

Provider-routing docs:

> **Strict mode** (`require_parameters: true`): The system only sends requests to providers explicitly supporting every parameter you've specified.

So attaching `reasoning` unconditionally does narrow routing, and `"No endpoints found that support..."` is the real failure mode. Still open.

### ✅ #7 confirmed (already fixed in `3557983`)

Worth recording *why* the fix mattered — the reasoning docs are explicit:

> An empty reasoning object enables reasoning at default parameters, typically "medium" effort with no exclusions.

So the pre-fix `reasoning: {}` would have silently switched reasoning **on** at medium effort. The `if (!("effort" in thinking)) return null;` guard is exactly right.

---

## Two new findings from the docs

### N1 — `--thinking <N>` bypasses OpenRouter's `reasoning.max_tokens` bounds (Medium)

Docs, on `reasoning.max_tokens`:

> Supported by: Gemini thinking models, Anthropic Claude reasoning models (**minimum 1024 tokens; maximum 128,000 tokens**), some Alibaba Qwen models

The level-based path is safe — `genericScale` is 1024/4096/8192/24576, all in range. But `resolveThinking`'s explicit-budget branch returns the user's number raw:

```ts
if (intent.kind === "budget") {
  const tokenField = tokenFieldFor(allowedFields);
  return tokenField === "max_tokens" ? { max_tokens: intent.tokens } : { budget_tokens: intent.tokens };
}
```

and `clampToOutputRoom` deliberately no-ops for `max_tokens` (`if (context.field !== "budget_tokens") return target;`). Since the openrouter profile is `fields: ["max_tokens"]`, every explicit budget lands unclamped in `reasoning.max_tokens`.

`parseThinking` accepts any positive integer, so `--thinking 500` against `openrouter/anthropic/claude-sonnet-4.6` sends `reasoning: {max_tokens: 500}` — below the documented 1024 floor — and the provider rejects the turn. `--thinking 200000` breaches the ceiling the same way.

### N2 — `reasoning_details` replay order is never sorted (Low)

Docs state a hard constraint:

> The entire sequence of consecutive reasoning blocks **must match the outputs generated by the model** during the original request; you cannot rearrange or modify the sequence.

`streamOpenRouter` serializes with `[...reasoningDetails.values()]`, which is `Map` **insertion** order — first-seen-in-stream, never sorted by `index`. In practice deltas arrive in order so this usually holds, but the new keying mixes three namespaces (`id:` / `index:` / `position:`) within a single turn, so a provider that starts emitting `id` partway through can interleave entries relative to the model's original sequence. Sorting by numeric `index` before `JSON.stringify` is cheap insurance against a constraint the docs call out explicitly.

### N3 — nit, latent only

`effort: thinking.effort === "max" ? "high" : thinking.effort` downgrades the top tier, but the docs list `"max"` (and `"xhigh"`) as valid effort values allocating ~95% of tokens. Currently unreachable because the openrouter profile forces `max_tokens`, so this only bites if an `effort` field is ever allowed for the provider.

---

**Verdict unchanged: ready to merge.** N1 and N2 are follow-ups, not blockers.

PR #6, opened 2026-08-17
Add custom sub-agent registry with background execution and time budgets

muhammadbalawal:
Sub-agents shared one hardcoded system prompt and inherited the parent's model and tool set wholesale. This makes them definable per project, lets them run past the turn that spawned them, and gives them wall-clock budgets.

## Defining an agent

`.backboard/agents/researcher.md` — frontmatter configures, body becomes the system prompt:

```markdown
---
description: Deep-dives one question, read-only.
tools: [read, grep, glob]
model: anthropic/claude-opus-5
maxRounds: 30
---

You are a read-only research sub-agent. Never modify files.
Report findings as file paths with a one-line summary each.
```

The parent then calls `Agent(subagent_type: "researcher", prompt: "…")`. Identity comes from the file; the task comes from the parent at call time.

## What's in here

**Agent registry** — discovery from `.backboard/agents/*.md` (project) and `~/.backboard/agents` (user), following the existing MCP/hooks config split. Per-agent `tools`, `disallowedTools`, `model`, `maxRounds`, `timeoutMs`, `background`. `subagent_type` becomes a dynamic enum built from the catalog. Invalid files are skipped with a startup warning, never fatal. `worker` and `rlm` remain built-ins, so behaviour is unchanged with no agent files present.

**Background execution** — `background: true` runs an agent past its spawning turn under a supervisor whose cancellation signal is deliberately unlinked from the turn, so cancelling the foreground no longer kills it. Reports arrive through a new `later` tier on the controller's submit queue, selected at drain time so user input can never be starved. Running agents show in the status bar.

**Time budgets** — `RLMBudget` extracted into a shared `RunBudget` and applied to worker agents. On expiry the agent gets one tool-less turn to report what it established rather than having its work discarded, returning `status: timed_out`.

## Fixes found along the way

- **Concurrent spawns were unbounded** — 20 `Agent` calls in one message meant 20 live sessions. Now capped, with a separate pool for background runs.
- **Deadlock risk** — a nested chain holds one permit per level, so `maxConcurrent < maxDepth` could deadlock against itself. Rejected at construction; a test exercises the tight boundary where permits exactly equal depth.
- **`AgentCatalog` was last-write-wins** — precedence only worked because discovery de-duplicated first. The catalog now keeps the first definition itself.
- **Agent reports rendered shredded** — a raw 4000-char slice that the transcript then truncated mid-line. Now routed through `buildOutputPreview` like every other tool.
- **Headless mode promised reports it would cancel** — `--print` exits after one prompt, so background spawns there now run inline.
- **New threads leaked stale reports** — an agent spawned for a discarded conversation would report into the fresh one.

## Design decisions worth reviewing

- **Background is top-level only.** A nested background spawn would outlive the sub-agent that owns it, leaving no one to receive its report. Deeper spawns run inline regardless of the flag.
- **Nesting is opt-in.** An agent that pins `tools` must list `agent` explicitly to keep spawning.
- **`/undo` does not cover background agents.** They outlive their turn's checkpoint, so journaling edits there would write into an already-finalized checkpoint. Capture is disabled instead; the README says to prefer read-only tools for them.
- **Agent names are matched exactly, not normalized.** Lowercasing `Researcher.md` would collide with `researcher.md`, so the rule is stated in the validation error instead.

## Testing

`bun run validate` clean — 1416 pass, 0 fail. Eight new test files covering discovery and parsing, tool-allowlist policy, the semaphore, worker and RLM timeouts, submit-queue priority, the background supervisor, and an end-to-end flow that drives a real `AgentController` + supervisor + `AgentTool` and asserts the spawning turn returns while the agent is still running, that a second user turn is not blocked, and that the report arrives as its own turn afterwards.

Not covered by tests: live Ink rendering, and a model choosing a background agent on its own. Both were exercised manually.

## Review note

The three areas interleave within `AgentTool.tsx`, `SubAgentRunner.ts`, `cli.tsx` and `Store.ts`, so splitting them into separate commits would have produced commits that don't compile. Kept as one commit with the areas separated in the message.


src/tools/AgentTool.tsx
```diff
@@ -159,33 +193,83 @@ export class AgentTool extends Tool<AgentToolInput, AgentToolOutput> {
...
+			},
+			report,
+			`Running in background (${snapshot.id})`,
+		);
+	}
+
+	private async runRlm(run: AgentRun): Promise<ToolResult<AgentToolOutput>> {
+		const { input, ctx, definition, trace, tracePath } = run;
+		const result = await this.gated(run, () =>
+			this.deps.createRLMLoop().run({
```
0vp: [P1] Apply custom RLM configuration

Custom `mode: rlm` agents ignore their configured `systemPrompt`, `model`, and `maxRounds`. `runRlm()` creates the globally configured RLM loop and only passes the definition's timeout, so the selected custom agent does not actually run as configured.

src/core/agent/SubAgentRunner.ts
```diff
@@ -84,21 +113,34 @@ export class SubAgentRunner {
...
+		const scoped = params.parentTurnId
 			? parentRecorder?.scopedToTurn(params.parentTurnId)
 			: undefined;
+		const revocable = scoped ? revocableRecorder(scoped) : undefined;
+		const checkpoints = revocable?.recorder;
+
+		// A budget bounds how long someone waits. With a handoff, or once the
+		// chain is already backgrounded, nobody is: expiry must not abort.
+		const canHandOff = params.onDeadline !== undefined;
+		const noWaiter = params.parentInBackground === true;
```
0vp, line 125: [P1] Treat explicit background agents as backgrounded

Agents launched with `background: true` receive `chainInBackground: true`, but timeout behavior checks only `parentInBackground`. Their configured timeout therefore aborts them instead of allowing them to continue, contrary to the documented behavior for an already-backgrounded chain.

src/core/checkpoints/CheckpointStore.ts
```diff
@@ -86,6 +86,41 @@ export interface CheckpointRecorder {
...
+	return {
+		revoke: () => {
+			live = false;
+		},
+		recorder: {
+			recordPreImage: (...args) =>
+				live ? recorder.recordPreImage(...args) : noop(),
+			recordPostImage: (...args) =>
+				live ? recorder.recordPostImage(...args) : noop(),
+			revertToolCall: (...args) =>
```
0vp: [P1] Preserve ApplyPatch rollback after handoff

Handoff revokes `revertToolCall`. If the deadline fires while ApplyPatch is between filesystem operations and a later operation fails, ApplyPatch attempts its atomic rollback, but rollback is now a no-op. Earlier file mutations remain despite the tool reporting failure.

src/core/checkpoints/CheckpointStore.ts
```diff
@@ -86,6 +86,41 @@ export interface CheckpointRecorder {
...
+				live ? recorder.revertToolCall(...args) : noop(),
+			beginShellCapture: (...args) =>
+				live ? recorder.beginShellCapture(...args) : noop(),
+			endShellCapture: (...args) =>
+				live ? recorder.endShellCapture(...args) : noop(),
+			captureWarning: () => (live ? recorder.captureWarning() : null),
+			scopedToTurn: (turnId) => recorder.scopedToTurn(turnId),
```
0vp: [P1] Keep nested checkpoint recorders revocable

`scopedToTurn()` returns a recorder outside the revocation wrapper. Nested agents can therefore retain checkpoint access after their top-level parent moves into the background, allowing edits to be journaled against an already-finalized foreground checkpoint.

src/core/agent/AgentController.ts
```diff
@@ -611,6 +631,10 @@ export class AgentController {
 	/** Cancels any active turn and starts a fresh Backboard thread. */
 	newThread(): void {
 		this.cancel({ clearQueue: true });
+		// Background agents were spawned to serve the discarded conversation;
+		// letting them report into the new thread would reference context that
+		// no longer exists.
+		this.deps.backgroundSupervisor?.cancelAll();
```
0vp, line 641: [P1] Cancel background runs when resuming a session

Background runs are cancelled by `newThread()`, but not by `hydrateSession()`. If the user resumes another session while an agent is running, the old agent's eventual notifier submits that report into the newly resumed conversation.

0vp (commented):
Reviewed the custom-agent registry, timeout handoff, background lifecycle, and checkpoint integration. The overall direction looks good, but I found five correctness issues that should be addressed.

src/core/checkpoints/CheckpointStore.ts
0vp: The updated `revertToolCall` fixes the original path, but a focused reproduction still leaves a partial patch through a different path. After revocation, `recordPreImage()` returns success without capturing anything even when ApplyPatch passes `requireRevertible: true`. A two-file patch then writes the first file, fails on the second operation, and rollback has no pre-images, so the first file remains. Please reject revoked `requireRevertible` captures before mutation or preserve an independent atomic rollback surface.

src/core/agent/AgentController.ts
0vp: `cancelAll()` closes the original straightforward case, but the session race is still reproducible. A background run can finish after the UI's final `hasActiveWork` check; `finish()` removes it from the supervisor and its notifier starts the report turn. `hydrateSession()` then calls `cancelAll()`, which has nothing left to cancel, and the active stale turn appends its assistant message to the resumed session. The reproduction retained the new thread id but produced `[new session baseline, stale report processed]`. Hydration also needs to cancel active/queued controller work, or report submission and session switching need an atomic barrier.

src/core/checkpoints/CheckpointStore.ts
```diff
@@ -86,6 +86,48 @@ export interface CheckpointRecorder {
...
+ * Wraps a recorder so journaling can be switched off after contexts holding it
+ * have already been copied. Used when a run outlives the turn it was scoped to.
+ */
+export function revocableRecorder(
+	recorder: CheckpointRecorder,
+): RevocableRecorder {
+	const gate = { live: true };
+	return {
+		revoke: () => {
+			gate.live = false;
```
0vp, line 106: [P1] Drain in-flight captures before completing handoff

A recorder call that observes `gate.live` before revocation can continue through asynchronous file/blob I/O and append after the parent turn is finalized, while its later post-image is suppressed. I reproduced this by delaying `recordPreImage()`, revoking and finalizing the turn, then releasing the capture: the finalized checkpoint included the background file, and `/undo` restored it from the completed background result back to its old content. Revocation needs to wait for active captures or prevent an in-flight capture from committing after the gate closes.

0vp (commented):
Follow-up verification found one additional checkpoint race. The other two remaining issues are documented with reproductions in their existing review threads.

src/core/checkpoints/CheckpointStore.ts
0vp: The new `mayJournal` checks fix revocation during `lstat`/blob reads, but the same issue remains if revocation happens during the final `journal.flush()`. I reproduced it by delaying that flush after the pre-image was appended, then revoking and finalizing the turn. The background post-image was suppressed, the finalized checkpoint retained the pre-image, and `/undo` restored the completed background file to its old content. Please re-check after the flush before returning (throwing for `requireRevertible`), or make revocation wait for active captures to drain.

src/core/agent/AgentController.ts
0vp: The new `hydrateSession()` cancellation fixes a report turn that was already active when hydration begins, but cancellation still happens after `SessionLifecycle.activate()` rotates checkpoint and event-log roots. I reproduced the remaining window by activating the replacement `ClientEventLog`, starting a stale background report that invoked a side-effect tool, and then hydrating: the tool executed and the replacement log contained the stale `turn:start` before cancellation. Background/controller work needs to be cancelled before `onResumeLocalSession`, `onResumeRemoteSession`, or `onNewSession` activates the durable replacement session.

src/core/agents/parse.ts
```diff
@@ -0,0 +1,148 @@
...
+	if (tools === INVALID) return skip("tools must be a list of strings.");
+
+	const disallowedTools = stringList(data.disallowedTools);
+	if (disallowedTools === INVALID) {
+		return skip("disallowedTools must be a list of strings.");
+	}
+
+	const modelValue = stringValue(data.model);
+	const model =
+		modelValue && modelValue !== "inherit" ? parseModel(modelValue) : undefined;
```
0vp, line 72: [P2] Reject malformed model references during discovery

Values such as `model: /` are accepted as `{ provider: "", model: "" }`, so discovery treats the agent as valid instead of producing the documented startup warning. Invoking it then sends an invalid model reference downstream and fails at runtime. Require non-empty provider and model components for qualified references, apart from the explicit `inherit` value.

0vp (commented):
Full follow-up review found one additional configuration-validation issue. Two remaining orchestration races are documented with reproductions in their existing threads.

PR #11, opened 2026-08-18
Add session resume support

0vp:
## Summary
- add `--resume` startup support and `/resume`, `/session`, and `/sessions` commands for Backboard, BYOK, and local sessions
- generate reusable exit hints that preserve original CLI flags, replace stale resume IDs, and only retain `--cwd` when explicitly supplied
- add a user-level resume index so local IDs reopen their original workspace from any directory without triggering SSO
- improve the session picker with updated time, compact IDs, message counts, and full-ID search
- make resume activation safer with self-resume protection, direct remote hydration, best-effort index persistence, and transactional session-log rotation
- share ID, shell quoting, and atomic private-file helpers across the implementation

## Validation
- `bun run validate`
  - 1,385 passed
  - 3 skipped (`cmd.exe` tests unavailable on macOS)
  - 0 failed
- interactively verified BYOK and local-only resume from arbitrary directories
- interactively verified `/sessions`, `/resume`, direct session switching, self-resume, read-only index behavior, and exit hints with and without explicit `--cwd`
- interactively verified direct resume of a real Backboard thread through active SSO from `/tmp`


src/ui/utils/resumeSession.ts
```diff
@@ -0,0 +1,155 @@
...
+
+	const threads = await client.listThreads();
+	const listed = threads.find(
+		(thread) =>
+			thread.thread_id === normalized ||
+			thread.metadata_?.[BYOK_SESSION_ID_METADATA_KEY] === normalized,
+	);
+	if (listed) return hydrateResumeTarget(client, listed);
+
+	if (
```
muhammadbalawal: **Critical — data loss (verified by repro test).** Resuming a Backboard-backed session by its local `sess_` id silently wipes the conversation. Only BYOK threads carry `backboard_session_id` metadata, so the `listThreads()` scan above never matches a remote thread; control falls into this branch and returns `{thread: null, messages: []}`. `activateResumeTarget` then hits `if (!thread) controller.newThread()` → session reset, and the UI reports `Resumed session: sess_… · 0 messages`. Repro: remote-only thread + existing local session dir → `newThread()` fires, `hydrateSession()` never does.

src/entrypoints/cli.tsx
```diff
@@ -143,15 +203,23 @@ async function main(): Promise<void> {
...
+	await registerResumeIds({
 		cwd: config.cwd,
-		model: config.modelString,
-		profile: config.profile.name,
-	});
+		sessionId,
+	}).catch(() => undefined);
 
 	const bus = new EventBus();
 	const clientLog = new ClientEventLog(sessionId, store.paths.clientLog);
```
muhammadbalawal, line 254: **Critical — log corruption (verified).** A resumed launch reuses the existing session dir, but `ClientEventLog`/`ServerEventLog` constructors start `sequence = 0` — only `activate()` calls `nextJsonlSequence`, and `sessionLifecycle.resume()` early-returns when `root === currentSessionRoot`, so the repair never runs. The resumed run appends records numbered 0,1,2… into a client.jsonl that already holds 0..400. Verified: two runs on one file produce sequences `[0,1,2,0,1]`; any consumer ordering/deduping by `sequence` sees collisions.

src/ui/utils/resumeSession.ts
```diff
@@ -0,0 +1,155 @@
...
+		return;
+	}
+
+	if (thread.metadata_?.[BYOK_THREAD_METADATA_KEY] === true) {
+		const provider = thread.metadata_?.model_provider;
+		const model = thread.metadata_?.model_name;
+		if (typeof provider === "string" && typeof model === "string") {
+			options.config.setModel({ provider, model });
+			options.controller.setModelContextLimit(null);
+			await options.config.saveRuntimeSelection().catch((error) => {
```
muhammadbalawal: **High — silent global side effect (verified).** This runs on every BYOK resume, including non-interactive `--print … --resume byok_…`, and `saveRuntimeSelection()` persists `model: currentModel` into `~/.backboard/config.json` whenever `--model` wasn't passed. A scripted/CI resume permanently rewrites the user's global default model with no prompt — and per the `onWarning` issue in cli.tsx, no visible warning if the write fails.

src/entrypoints/cli.tsx
```diff
@@ -372,6 +451,31 @@ async function main(): Promise<void> {
...
+			initialResumeTarget = await resolveResumeTarget(
+				client,
+				config.cwd,
+				config.flags.resume,
+			);
+			await activateResumeTarget(initialResumeTarget, {
+				config,
+				controller,
+				onResumeLocalSession: (id) => sessionLifecycle.resume(id),
+				onWarning: (message) => bus.emit({ type: "system:warning", message }),
```
muhammadbalawal: **High — warning silently dropped (verified).** This emit runs before `render()` and before `runHeadless` attaches its `system:warning` handler; `EventBus` has no replay, so the message goes nowhere but the JSONL log. The comment at line ~290 of this same function documents the invariant: pre-render warnings must go through `startupWarnings`, not `bus.emit`. Route `onWarning` there.

src/entrypoints/cli.tsx
```diff
@@ -112,10 +130,52 @@ async function main(): Promise<void> {
...
 	}
 
+	const requestedResume = earlyFlags.resume?.trim();
+	const indexedResume = requestedResume
+		? await lookupResumeEntry(requestedResume)
+		: null;
+	const resumeCwd = resolvePath(
+		earlyFlags.cwd ?? indexedResume?.cwd ?? process.cwd(),
+	);
+	const runtimeArgv =
```
muhammadbalawal, line 154: **High — silent cwd relocation.** When `--cwd` wasn't passed and the id is in the index, this silently moves the entire run to the indexed cwd: file tools, permission rules, checkpoints, MCP/hook project config, and LSP all operate on the other workspace with nothing in the output saying so. Since `registerResumeIds` also indexes remote thread ids at exit, a pasted id from another project is enough to trigger it. At minimum print a notice that the cwd changed.

src/core/session/ResumeIndex.ts
```diff
@@ -0,0 +1,120 @@
...
+	input: {
+		cwd: string;
+		sessionId: string;
+		threadId?: string | null;
+	},
+	homeDir?: string,
+): Promise<void> {
+	const path = resumeIndexPath(homeDir);
+	await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+	const lease = await acquireFileLease(`${path}${RESUME_INDEX_LEASE_SUFFIX}`, {
```
muhammadbalawal, line 51: **Medium — lease not re-entrant (verified).** `leaseOwnerIsAlive` is true for the current live PID, so a second acquire from the same process can never reclaim: it spins on `retryMs` for the full `timeoutMs` and throws. cli.tsx fires `void registerResumeIds(...)` from the `session:thread` handler while the exit path `await`s another — quitting right after a thread event stalls exit for up to 5s, and the losing call's write is silently discarded by `.catch(() => undefined)`. Verified: second in-process acquire burns the full timeout and throws.

src/core/session/ResumeIndex.ts
```diff
@@ -0,0 +1,120 @@
...
+	});
+	try {
+		const file = await readResumeIndex(homeDir);
+		const entry: ResumeIndexEntry = {
+			cwd: resolve(input.cwd),
+			sessionId: input.sessionId,
+			...(input.threadId ? { threadId: input.threadId } : {}),
+			updatedAt: new Date().toISOString(),
+		};
+		file.entries[input.sessionId] = entry;
```
muhammadbalawal, line 84: **Medium — unbounded growth (verified).** These are pure inserts — no TTL, cap, or eviction, and `updatedAt` is written but never consulted. cli.tsx registers on *every* launch (resumed or not), so the index grows by up to 2 entries per launch forever, and each launch pays mkdir + lease + full JSON parse + full rewrite before the UI appears. Verified: 5 launches → 10 entries, none evicted.

src/core/session/ResumeIndex.ts
```diff
@@ -0,0 +1,120 @@
...
+		timeoutMs: 5_000,
+		retryMs: 50,
+		invalidOwnerStaleMs: 5_000,
+	});
+	try {
+		const file = await readResumeIndex(homeDir);
+		const entry: ResumeIndexEntry = {
+			cwd: resolve(input.cwd),
+			sessionId: input.sessionId,
+			...(input.threadId ? { threadId: input.threadId } : {}),
```
muhammadbalawal: **Medium — dead field, incorrectly maintained (verified).** No consumer reads `entry.threadId` (cli.tsx only reads `.cwd` and `.sessionId`), and the every-launch `registerResumeIds({cwd, sessionId})` call replaces `entries[sessionId]` wholesale, clobbering the threadId recorded at the previous exit — while the threadId-keyed alias entry keeps the stale copy, so the two disagree. Either use the field or drop it.

src/core/session/ResumeBootstrap.ts
```diff
@@ -0,0 +1,100 @@
...
+	id: string | undefined,
+	sessionIdHint?: string,
+): Promise<LocalResumeBootstrap | null> {
+	const normalized = id?.trim();
+	if (!normalized || !isLocalResumeId(normalized)) return null;
+	const conversationStore = new ByokConversationStore(cwd);
+	const likelySessionId = isSessionId(normalized) ? normalized : sessionIdHint;
+	const directConversation = likelySessionId
+		? await conversationStore.getAtSessionRoot(
+				qSessionDir(cwd, likelySessionId),
```
muhammadbalawal, line 34: **Medium — path traversal (verified).** `sessionIdHint` comes from the on-disk index, where `isResumeIndexEntry` only checks `typeof value.sessionId === "string"` — no shape validation. A corrupted or hand-edited `session-index.json` with `"sessionId": "../../../../etc"` flows straight into `qSessionDir` and escapes the workspace (verified: resolves outside cwd). One-line fix: ignore hints that fail `isSessionId`.

src/utils/id.ts
```diff
@@ -9,6 +9,23 @@ export function shortId(prefix = ""): string {
 	return prefix ? `${prefix}_${id}` : id;
 }
 
+export function isShortId(value: string, prefix: string): boolean {
+	const expectedPrefix = `${prefix}_`;
+	return (
+		value.length === expectedPrefix.length + 8 &&
+		value.toLowerCase().startsWith(expectedPrefix.toLowerCase()) &&
```
muhammadbalawal: **Medium — case-insensitive validation vs case-sensitive lookup (verified).** `isShortId` accepts uppercase (`isSessionId("SESS_1234ABCD") === true`) but callers use the id verbatim as a directory name, and generated ids are always lowercase. The same command resolves on APFS (macOS) and fails with `Local session … was not found` on ext4 (Linux). Drop the case-insensitivity or normalize before use.

src/entrypoints/cli.tsx
```diff
@@ -372,6 +451,31 @@ async function main(): Promise<void> {
 			await sessionLifecycle.dispose();
 		}
 	};
+	let initialResumeTarget: ResumeTarget | undefined;
+	if (config.flags.resume !== undefined) {
```
muhammadbalawal, line 495: **Low — empty `--resume` fails late with a stack trace (verified).** `parseFlags(["--resume"]).resume === ""` (also `--resume --lsp`, since flag-like values are rejected): falsy for the early `requestedResume` guards, but `!== undefined` here — so `backboard --resume` skips every early check, does full startup (MCP init, LSP, session lease), then throws `A session ID is required.` plus a stack trace. Reject the empty value up front.

src/entrypoints/cli.tsx
```diff
@@ -112,10 +130,52 @@ async function main(): Promise<void> {
...
+		earlyFlags.cwd === undefined && indexedResume
+			? [...argv, "--cwd", resumeCwd]
+			: argv;
+	const localResume = await resolveLocalResumeBootstrap(
+		resumeCwd,
+		earlyFlags.resume,
+		indexedResume?.sessionId,
+	);
+	if (requestedResume && isLocalResumeId(requestedResume) && !localResume) {
+		throw new Error(
```
muhammadbalawal: **Low — user-input errors surface as stack traces.** This throw (and `Backboard sign-in is required…` below, and the rethrown `Session … was not found.`) propagates to `main().catch` → `formatFatalError`, which returns `err.stack ?? err.message` — a full JS stack trace for a typo'd session id. Every other user-facing error in this file goes through `errorMessage(err)` + `process.exit(1)`; these should too.

src/core/session/ResumeBootstrap.ts
```diff
@@ -0,0 +1,100 @@
...
+				isByokThreadId(normalized) ? normalized : undefined,
+			)
+		: null;
+	const conversation =
+		directConversation ??
+		(await conversationStore.list()).find(
```
muhammadbalawal: **Perf — workspace parsed three times on the blocking startup path.** For a local BYOK resume: (1) this `list()` readdirs and JSON-parses every conversation.json; (2) `resolveResumeTarget` → `ByokClient.listThreads()` → `conversationStore.list()` does it again; (3) `hydrateResumeTarget` → `getThread()` → `get()` → `listConversations()` a third time. The listThreads result already contains the hydrated thread for BYOK, so pass (3) is pure waste — and all of it runs before the first frame renders.

src/entrypoints/cli.tsx
```diff
@@ -112,10 +130,52 @@ async function main(): Promise<void> {
...
+			`Local session "${requestedResume}" was not found in ${resumeCwd}. Run the command from its original workspace or pass --cwd <workspace>.`,
+		);
+	}
+	if (isRemoteResumeId(requestedResume) && resolveAuth().backboard === null) {
+		if (
+			earlyFlags.print !== undefined ||
```
muhammadbalawal: **Low — headless detection diverges from the canonical check (verified).** This compares the raw flag (`earlyFlags.format === "json"`) while line ~284 uses parsed `config.format` (`parseOutputFormat` trims + lowercases). So `backboard --format JSON --resume thread_… ` without credentials takes the interactive branch and launches the device-code login inside a machine-readable run, instead of the intended error. `shouldShowAuthScreen` has the same raw compare — worth one shared predicate instead of a third copy.

src/entrypoints/cli.tsx
```diff
@@ -435,11 +540,22 @@ async function main(): Promise<void> {
 		maxFps: 12,
 	});
 
+	let activeId = controller.threadId ?? sessionLifecycle.current().sessionId;
```
muhammadbalawal: **Nit — dead initializer.** `activeId` is unconditionally reassigned in the `finally` before its only read (the `try` only awaits `waitUntilExit`), so this render-time computation does nothing. `let activeId: string;` — or just compute it once in the `finally` — says the same thing without implying the pre-UI value matters.

src/entrypoints/cli.tsx
```diff
@@ -213,7 +296,7 @@ async function main(): Promise<void> {
 	// Both --print and --format json run through runHeadless, which has no UI to
 	// answer an input:request. Anything that asks there would hang forever, so
 	// the gate must know there is nobody to prompt.
-	const headless = config.flags.print !== undefined || config.format === "json";
+	const headless = isHeadlessInvocation(config.flags, config.format);
```
muhammadbalawal: `isHeadlessInvocation` adds `|| !stdinIsTTY || !stdoutIsTTY` (`resumePreflight.ts:37-42`), but the CLI does **not** go headless for those — neither early return at 498 nor 510 fires, so `render(app)` runs at 554 and the user has a working UI with `interactive: false`.

That flag feeds two different places, and the first is a permission bypass. Measured with `--permission-mode manual`, stdin a TTY, only stdout differing, each row a fresh process:

```
                           stdout=TTY      stdout=piped
echo hello                 PROMPTS    ->   ALLOWED, runs silently
rm -rf /tmp/somedir        PROMPTS    ->   DENIED
curl https://example.com   PROMPTS    ->   DENIED
git push --force           PROMPTS    ->   DENIED
```

`PermissionEngine.ts:49` gates `strictManual` on `pctx.interactive`, so `interactive: false` switches manual mode's blanket prompting **off** and ordinary commands fall through to `tool.checkPermissions` and just run. Whatever is still classified `ask` then hits `resolveToolPermission.ts:82` and is denied.

Net: `backboard --permission-mode manual | tee run.log` gives weaker gating than `auto` while also breaking the dangerous-command path. The `stdout=TTY` column is exactly what main computes for that same piped invocation (`main` cli.tsx:216 = `flags.print !== undefined || format === "json"`), so all four arrows are regressions.

Worth noting this is collateral, not part of the feature — nothing about session resume requires changing the permission gate. The preflight needs "should I force a login before Config exists"; the permission context needs "is there a UI attached", which here is just "did we reach `render()`". Two different questions sharing one helper.

src/ui/utils/resumeSession.ts
```diff
@@ -0,0 +1,177 @@
...
+			indexedResume?.threadId &&
+			resolve(indexedResume.cwd) === resolve(cwd)
+		) {
+			try {
+				const target = resumeTargetFromHydratedThread(
+					await client.getThread(indexedResume.threadId),
+				);
+				return { ...target, localSessionId: normalized };
+			} catch (error) {
+				if (!(error instanceof BackboardError) || error.status !== 404) {
```
muhammadbalawal: Both catch branches throw, so the local-session fallback 30 lines below is unreachable from here:

```
isSessionId && indexedResume.threadId && same cwd
  try getThread(threadId)
    catch non-BackboardError -> throw            (BYOK conversation file deleted)
          404                -> throw CliUserError (thread deleted server-side)
                                                   (network error is non-404 -> throw)

  never reached v
  line 84  isSessionId && isDirectory(qSessionDir(cwd, id)) -> resume with messages: []
```

So a local session whose directory is present and perfectly resumable dies because the index happens to carry a stale `threadId`. The same shape breaks offline resume: any network error resolving a remote threadId is non-404, so it rethrows instead of degrading to the local session.

Falling through to the existing fallback on any failure — rather than throwing — would keep the local session usable and matches what the code below already knows how to do.

src/entrypoints/cli.tsx
```diff
@@ -112,14 +136,62 @@ async function main(): Promise<void> {
...
+		throw new CliUserError(errorMessage(error));
+	}
+	const requestedResume = parseRequestedResume(earlyFlags.resume);
+	const indexedResume = requestedResume
+		? await lookupResumeEntry(requestedResume)
+		: null;
+	const resumeCwd = resolvePath(
+		earlyFlags.cwd ?? indexedResume?.cwd ?? process.cwd(),
+	);
+	const runtimeArgv =
```
muhammadbalawal, line 154: `indexedResume` comes from `lookupResumeEntry(requestedResume)` for **any** resume ID, including remote Backboard thread IDs, so this silently relocates the agent's working directory:

```ts
resumeCwd   = earlyFlags.cwd ?? indexedResume?.cwd ?? process.cwd()
runtimeArgv = !earlyFlags.cwd && indexedResume ? [...argv, "--cwd", resumeCwd] : argv
```

`cd ~/repo-b && backboard --resume thread_xyz` now runs every file read, edit, shell command, checkpoint and permission rule against `~/repo-a`, with nothing printed to say so. The `--cwd` reroute is defensible for local `sess_`/`byok_` IDs whose session dir lives in a specific workspace; for a remote thread it is a guess.

Related, same mapping: `cli.tsx:484` writes `entries[threadId] = {cwd: config.cwd, ...}` on exit, so inspecting a thread once from `/tmp` permanently repoints every future `--resume thread_xyz` at `/tmp` — including the command the exit hint prints. There is no way to see or correct the mapping.

At minimum, print the relocation when it happens. Better: only apply it for local resume IDs.

src/entrypoints/cli.tsx
```diff
@@ -112,14 +136,62 @@ async function main(): Promise<void> {
...
+		throw new CliUserError(
+			`Local session "${requestedResume}" was not found in ${resumeCwd}. Run the command from its original workspace or pass --cwd <workspace>.`,
+		);
+	}
+	const remoteResumeId = isRemoteResumeId(requestedResume)
+		? requestedResume
+		: indexedResume?.threadId && !isByokThreadId(indexedResume.threadId)
+			? indexedResume.threadId
+			: undefined;
+	if (remoteResumeId && resolveAuth().backboard === null) {
```
muhammadbalawal, line 175: `isRemoteResumeId` is just the negation of the local test (`ResumeBootstrap.ts:62-65`):

```ts
export function isRemoteResumeId(id) { return Boolean(id?.trim() && !isLocalResumeId(id)); }
export function isLocalResumeId(id)  { return isByokThreadId(id) || isSessionId(id); }
```

and `parseRequestedResume` only rejects malformed IDs that already *start with* `sess_`/`byok_`. So anything else — a typo, a session title pasted instead of an ID — classifies as a Backboard thread. For a user with only a provider key, `resolveAuth().backboard === null` and the invocation is interactive, so line 177 launches the device-code SSO flow.

`backboard --resume my-old-session` should say `Session "my-old-session" was not found.`, not force an SSO login the user never asked for (or make them abort with exit code 1). The ID is never checked against the local BYOK store first.

src/entrypoints/cli.tsx
```diff
@@ -112,14 +136,62 @@ async function main(): Promise<void> {
...
+		}
+	}
 	let config: Config;
 	while (true) {
 		try {
-			config = new Config({ argv });
+			config = new Config({
+				argv: runtimeArgv,
+				resumeModel: localResume?.model,
+				allowUnauthenticatedResume: localResume !== null,
```
muhammadbalawal, line 193: `allowUnauthenticatedResume: localResume !== null` is also true for `kind: "session"` bootstraps, not just BYOK ones. A session directory with valid `meta.json`, no BYOK conversation, no Backboard token and no provider keys then takes this path:

- `remoteResumeId` is undefined, so no login is forced
- `Config` skips the `NO_CREDENTIALS_MESSAGE` throw because unauthenticated resume is allowed
- `ClientRouter.listThreads` sees only the BYOK client and finds nothing
- `resolveResumeTarget` falls to the directory branch and returns `messages: []`

The user sees `Resumed session: sess_abcd1234` with zero messages and no indication they are unauthenticated. The first turn then fails with "No backend can serve …".

Either gate the allowance to bootstraps that actually carry local credentials, or surface a warning when a session resumes with no backend behind it.

src/config/Config.ts
```diff
@@ -120,7 +129,10 @@ export class Config {
 				: null;
 		this.currentModel = this.flags.model
```
muhammadbalawal, line 130: `this.flags.model` outranks `options.resumeModel` here, but `activateResumeTarget` overwrites it afterwards anyway (`resumeSession.ts:150`), so an explicit `--model` alongside `--resume` is applied and then silently thrown away.

```
backboard --model openai/gpt-5 --resume byok_deadbeef      (thread metadata: anthropic/claude-*)

Config.ts:130          flag wins            -> openai/gpt-5
                       client + model profile built for gpt-5
resumeSession.ts:150   setModel(metadata)   -> anthropic/claude-*      <- user's flag gone
                       setModelContextLimit(null)
```

No warning, and the model changes out from under an already-resolved model profile. Either honour the explicit flag and skip the metadata restore, or tell the user the flag was overridden.

src/ui/utils/resumeSession.ts
```diff
@@ -0,0 +1,177 @@
...
+	if (!thread) {
+		options.controller.newThread();
+		return;
+	}
+
+	if (thread.metadata_?.[BYOK_THREAD_METADATA_KEY] === true) {
+		const provider = thread.metadata_?.model_provider;
+		const model = thread.metadata_?.model_name;
+		if (typeof provider === "string" && typeof model === "string") {
+			options.config.setModel({ provider, model });
```
muhammadbalawal, line 153: The extraction from `App.tsx` dropped the persistence step. On main (`App.tsx:1049-1054`) this block was:

```ts
config.setModel({ provider, model });
agent.setModelLabel(`${provider}/${model}`);
controller.setModelContextLimit(null);
await config.saveRuntimeSelection().catch((error) => {
  agent.notice(`Session resumed, but saving its model failed: ${errorMessage(error)}`, "error");
});
```

Here it stops at `setModelContextLimit`. The only two remaining `saveRuntimeSelection()` calls in `App.tsx` are the model picker and the memory picker, so nothing persists the resumed model any more, and the "saving its model failed" diagnostic is gone entirely.

A BYOK bootstrap happens to survive because `Config` reads `options.resumeModel`, but any path that does not populate it — remote-thread resume, `/resume` followed by `/new`, or an unreadable conversation file at preflight — reverts to the persisted default on the next launch.

src/ui/utils/resumeSession.ts
```diff
@@ -0,0 +1,177 @@
...
+function resumeTargetFromHydratedThread(
+	hydrated: BackboardThread,
+): ResumeTarget {
+	const byok = hydrated.metadata_?.[BYOK_THREAD_METADATA_KEY] === true;
+	const localSessionId = byok
+		? hydrated.metadata_?.[BYOK_SESSION_ID_METADATA_KEY]
+		: null;
+	if (
+		byok &&
+		(typeof localSessionId !== "string" || !isSessionId(localSessionId))
```
muhammadbalawal, line 115: This tightened the check from main, which only required a non-empty string (`App.tsx:1037`):

```
main   typeof sessionId !== "string" || sessionId.length === 0   -> throw
here   typeof localSessionId !== "string" || !isSessionId(...)   -> throw
```

`isSessionId` means exactly `sess_` + 8 lowercase hex, so any BYOK conversation whose stored session ID came from a different or older generator is now permanently unresumable through both the `/sessions` picker and `--resume`. There is no migration and no fallback to the `BYOK_SESSION_ROOT_METADATA_KEY` that is also recorded.

The message compounds it: "Saved BYOK session is missing its local session id" when the ID is present, just shaped differently. If the stricter check is deliberate, the error should say the ID is unrecognised and name what was found.

PR #12, opened 2026-08-18
Improve interactive input rendering responsiveness

### LINKED ISSUE #8: Input lag while typing/deleting characters
## Description
When typing into Backboard R-CLI, there is a noticeable input delay, more than with other TUIs such as `codex` (uses ratatui) and `claude` (uses Ink, same as R-CLI). This is most noticeable while deleting characters, and I believe it hurts the user experience.

## Steps to Reproduce
1. Go into R-CLI
2. Start typing (write a huge buffer, something like `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`)
3. Press backspace and do not release (simulation of the worst-case scenario)

## Expected Behavior
Steps 2 and 3 shouldn't look choppy, and it does not in `codex` or `claude`.

## Actual Behavior
Both steps look very choppy (due to the low refresh rate set by default).

## Environment
- OS: Linux
- Version / commit: Tested on both `6129f64` as well as `3.0.3`

## Video to show the difference (30 fps video):
No code change (maxFps: 12):

https://github.com/user-attachments/assets/510757fc-9e93-4453-9743-476a914717af

Code change (maxFps: 120, looks much less choppy to my eyes than in the 30 fps video, my screen has a 120 Hz framerate):

https://github.com/user-attachments/assets/797e0c9c-73de-4a08-a566-82792542dcb7

## Additional Context
- I've found a solution to this problem. You can change 2 lines of code on `src/entrypoints/cli.tsx` to replace `maxFps` from `12` to a higher value such as `60` (close to `claude`) or `120` (close to `codex`).
- I have a PR ready, however, I have not opened a pull request yet as external contributions do not seem to be accepted yet. If you want, I can open the PR, however, this is also a fix you could handle on your own.
- If you run the pre-commit hook from a normal terminal, you will see a test failing. I've opened another issue to address this: #9 

It is possible that the choice of keeping FPS at 12 was done for performance reasons, but I do not think that this is necessary for performance reasons, and as other harnesses have their max FPS set to much higher, I think it is reasonable for Backboard to also do so.

0vp:
## Summary

- keep Ink's append-only static transcript subtree out of ordinary React updates so it no longer bypasses frame throttling
- preserve banner insertion/removal semantics across `/resume`, `/new`, transcript resets, and keyed `<Static>` remounts
- stabilize static transcript arrays so steady-state memoization is O(1)
- raise the interactive render ceiling from 12 FPS to a balanced 30 FPS default for the main app and authentication screen
- support a validated `BACKBOARD_MAX_FPS` override from 1 to 120, with sanitized warnings for malformed or clamped values
- suppress interactive-only warnings in headless modes and surface them on the auth screen
- document the FPS override in the README
- add real Ink overflow, banner-transition, environment, auth-warning, and state-identity regression coverage

## Why

Issue #8 reports choppy typing and held-Backspace behavior. The original 12 FPS ceiling contributes visible latency, but profiling also found that rerendering `StaticTranscript` marked Ink's internal static node dirty and took an immediate-render path that bypassed `maxFps` entirely.

Memoizing stable append-only inputs makes the throttle effective. Banner presence is included in the keyed static identity so `/resume` → `/new` and banner removal cannot lose output through Ink's retained static index. A 30 FPS default then lowers worst-case visual latency from roughly 84 ms to 34 ms without allowing the terminal-output volume of a 60 FPS default. Faster local terminals can opt into 60 or 120 FPS with `BACKBOARD_MAX_FPS`.

Large pastes continue to use compact previews, and image/file attachments retain their existing asynchronous chip paths.

## Validation

- `bun run validate`
  - 1,367 passed
  - 3 skipped
  - 0 failed
- `BACKBOARD_MAX_FPS=15 bun test tests/RenderOptions.test.ts`
  - 6 passed
- real Ink overflow regression verifies dynamic updates remain behind the configured frame throttle
- banner insertion/removal regressions verify no session card or transcript item is lost
- interactive `tuistory` smoke test with the 30 FPS default

Fixes #8

### DIFF
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -58,7 +58,10 @@ import { createDefaultTools } from "../tools/index.ts";
 import { App } from "../ui/App.tsx";
 import { AuthScreen } from "../ui/AuthScreen.tsx";
 import { CLEAR_VISIBLE_SCREEN } from "../ui/hooks/ResizeStabilizer.constants.ts";
-import { interactiveRenderOptions } from "../ui/renderOptions.ts";
+import {
+	interactiveRenderOptions,
+	resolveInteractiveRenderConfig,
+} from "../ui/renderOptions.ts";
 import { palette } from "../ui/theme/palette.ts";
 import { ThemeProvider } from "../ui/theme/ThemeProvider.tsx";
 import { detectTerminalBg } from "../ui/theme/terminalBg.ts";
@@ -229,10 +232,18 @@ async function main(): Promise<void> {
 					`Unknown --permission-mode "${config.flags.permissionMode}"; using ${permissions.mode} mode. Valid: ${PERMISSION_MODES.join(", ")}.`,
 				]
 			: [];
+	const interactiveRenderConfig = resolveInteractiveRenderConfig();
+	const renderWarnings = interactiveRenderConfig.warning
+		? [interactiveRenderConfig.warning]
+		: [];
 	// MCP init warnings (unset env vars, skipped/failed servers) are deliberately
 	// kept out of the startup surface — they cluttered every launch. Server status
 	// is still inspectable via /mcp; only the noisy startup emission is dropped.
-	const startupWarnings = [...hookConfig.warnings, ...permissionWarnings];
+	const startupWarnings = [
+		...hookConfig.warnings,
+		...permissionWarnings,
+		...renderWarnings,
+	];
 	for (const warning of startupWarnings) {
 		bus.emit({ type: "system:warning", message: warning });
 	}
@@ -433,7 +444,10 @@ async function main(): Promise<void> {
 	);
 	const instance = render(
 		app,
-		interactiveRenderOptions({ exitOnCtrlC: false }),
+		interactiveRenderOptions({
+			exitOnCtrlC: false,
+			maxFps: interactiveRenderConfig.maxFps,
+		}),
 	);
 
 	try {
@@ -463,7 +477,13 @@ async function runAuthScreen(): Promise<boolean> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, interactiveRenderOptions({ exitOnCtrlC: true }));
+	const instance = render(
+		app,
+		interactiveRenderOptions({
+			exitOnCtrlC: true,
+			maxFps: resolveInteractiveRenderConfig().maxFps,
+		}),
+	);
 	await instance.waitUntilExit();
 	if (didLogin) {
 		instance.clear();
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -58,6 +58,7 @@ import { createDefaultTools } from "../tools/index.ts";
 import { App } from "../ui/App.tsx";
 import { AuthScreen } from "../ui/AuthScreen.tsx";
 import { CLEAR_VISIBLE_SCREEN } from "../ui/hooks/ResizeStabilizer.constants.ts";
+import { interactiveRenderOptions } from "../ui/renderOptions.ts";
 import { palette } from "../ui/theme/palette.ts";
 import { ThemeProvider } from "../ui/theme/ThemeProvider.tsx";
 import { detectTerminalBg } from "../ui/theme/terminalBg.ts";
@@ -430,10 +431,7 @@ async function main(): Promise<void> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, {
-		exitOnCtrlC: false,
-		maxFps: 12,
-	});
+	const instance = render(app, interactiveRenderOptions(false));
 
 	try {
 		await instance.waitUntilExit();
@@ -462,10 +460,7 @@ async function runAuthScreen(): Promise<boolean> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, {
-		exitOnCtrlC: true,
-		maxFps: 12,
-	});
+	const instance = render(app, interactiveRenderOptions(true));
 	await instance.waitUntilExit();
 	if (didLogin) {
 		instance.clear();
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -59,7 +59,7 @@ import { App } from "../ui/App.tsx";
 import { AuthScreen } from "../ui/AuthScreen.tsx";
 import { CLEAR_VISIBLE_SCREEN } from "../ui/hooks/ResizeStabilizer.constants.ts";
 import {
-	interactiveRenderOptions,
+	type InteractiveRenderConfig,
 	resolveInteractiveRenderConfig,
 } from "../ui/renderOptions.ts";
 import { palette } from "../ui/theme/palette.ts";
@@ -115,6 +115,7 @@ async function main(): Promise<void> {
 		await runLogoutCommand();
 		return;
 	}
+	const interactiveRenderConfig = resolveInteractiveRenderConfig();
 
 	let config: Config;
 	while (true) {
@@ -124,7 +125,7 @@ async function main(): Promise<void> {
 		} catch (err) {
 			const error = err instanceof Error ? err : String(err);
 			if (shouldShowAuthScreen(error, earlyFlags)) {
-				const didLogin = await runAuthScreen();
+				const didLogin = await runAuthScreen(interactiveRenderConfig);
 				if (!didLogin) return;
 				continue;
 			}
@@ -232,10 +233,7 @@ async function main(): Promise<void> {
 					`Unknown --permission-mode "${config.flags.permissionMode}"; using ${permissions.mode} mode. Valid: ${PERMISSION_MODES.join(", ")}.`,
 				]
 			: [];
-	const interactiveRenderConfig = resolveInteractiveRenderConfig();
-	const renderWarnings = interactiveRenderConfig.warning
-		? [interactiveRenderConfig.warning]
-		: [];
+	const renderWarnings = headless ? [] : interactiveRenderConfig.warnings;
 	// MCP init warnings (unset env vars, skipped/failed servers) are deliberately
 	// kept out of the startup surface — they cluttered every launch. Server status
 	// is still inspectable via /mcp; only the noisy startup emission is dropped.
@@ -442,13 +440,10 @@ async function main(): Promise<void> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(
-		app,
-		interactiveRenderOptions({
-			exitOnCtrlC: false,
-			maxFps: interactiveRenderConfig.maxFps,
-		}),
-	);
+	const instance = render(app, {
+		exitOnCtrlC: false,
+		maxFps: interactiveRenderConfig.maxFps,
+	});
 
 	try {
 		await instance.waitUntilExit();
@@ -457,7 +452,9 @@ async function main(): Promise<void> {
 	}
 }
 
-async function runAuthScreen(): Promise<boolean> {
+async function runAuthScreen(
+	interactiveRenderConfig: InteractiveRenderConfig,
+): Promise<boolean> {
 	let didLogin = false;
 	const { hex: terminalBg } = await detectTerminalBg(palette.bg);
 	const uiTheme = createTheme(terminalBg);
@@ -466,6 +463,7 @@ async function runAuthScreen(): Promise<boolean> {
 		<ThemeProvider value={uiTheme}>
 			<AuthScreen
 				keys={new ProviderKeyController()}
+				warnings={interactiveRenderConfig.warnings}
 				onKeySaved={() => {
 					didLogin = true;
 				}}
@@ -477,13 +475,10 @@ async function runAuthScreen(): Promise<boolean> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(
-		app,
-		interactiveRenderOptions({
-			exitOnCtrlC: true,
-			maxFps: resolveInteractiveRenderConfig().maxFps,
-		}),
-	);
+	const instance = render(app, {
+		exitOnCtrlC: true,
+		maxFps: interactiveRenderConfig.maxFps,
+	});
 	await instance.waitUntilExit();
 	if (didLogin) {
 		instance.clear();
--- a/src/entrypoints/cli.tsx
+++ b/src/entrypoints/cli.tsx
@@ -431,7 +431,10 @@ async function main(): Promise<void> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, interactiveRenderOptions(false));
+	const instance = render(
+		app,
+		interactiveRenderOptions({ exitOnCtrlC: false }),
+	);
 
 	try {
 		await instance.waitUntilExit();
@@ -460,7 +463,7 @@ async function runAuthScreen(): Promise<boolean> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, interactiveRenderOptions(true));
+	const instance = render(app, interactiveRenderOptions({ exitOnCtrlC: true }));
 	await instance.waitUntilExit();
 	if (didLogin) {
 		instance.clear();
--- a/src/ui/components/StaticTranscript.tsx
+++ b/src/ui/components/StaticTranscript.tsx
@@ -7,16 +7,18 @@ import type { RenderTranscriptItem, RunStatus } from "../../state/AppState.ts";
 import { Banner } from "./Banner.tsx";
 import { Item } from "./MessageList.tsx";
 
+export interface StaticTranscriptBanner {
+	status: RunStatus;
+	model: string;
+	cwd: string;
+	usage: UsageInfo;
+	update?: StartupUpdateInfo | null;
+}
+
 interface Props {
 	items: RenderTranscriptItem[];
 	generation: number;
-	banner: {
-		status: RunStatus;
-		model: string;
-		cwd: string;
-		usage: UsageInfo;
-		update?: StartupUpdateInfo | null;
-	} | null;
+	banner: StaticTranscriptBanner | null;
 }
 
 type StaticTranscriptItem =
@@ -54,7 +56,10 @@ function StaticTranscriptComponent({
 		: items;
 
 	return (
-		<Static key={generation} items={staticItems}>
+		<Static
+			key={`${generation}:${banner === null ? "plain" : "banner"}`}
+			items={staticItems}
+		>
 			{(item) =>
 				item.kind === "banner" ? (
 					<Banner
@@ -73,20 +78,4 @@ function StaticTranscriptComponent({
 	);
 }
 
-export function shouldReuseStaticTranscript(
-	previous: Props,
-	next: Props,
-): boolean {
-	if (previous.generation !== next.generation) return false;
-	if (previous.items.length !== next.items.length) return false;
-	// <Static> is append-only. Banner changes are picked up when generation
-	// changes and the transcript is intentionally reprinted. During ordinary
-	// App updates, avoid touching Ink's internal_static node because that
-	// bypasses maxFps and forces an immediate render.
-	return previous.items.every((item, index) => item === next.items[index]);
-}
-
-export const StaticTranscript = memo(
-	StaticTranscriptComponent,
-	shouldReuseStaticTranscript,
-);
+export const StaticTranscript = memo(StaticTranscriptComponent);
--- a/src/ui/components/StaticTranscript.tsx
+++ b/src/ui/components/StaticTranscript.tsx
@@ -1,5 +1,6 @@
 import { Static } from "ink";
 import type React from "react";
+import { memo } from "react";
 import type { UsageInfo } from "../../core/bus/events.ts";
 import type { StartupUpdateInfo } from "../../core/update/startupNotice.ts";
 import type { RenderTranscriptItem, RunStatus } from "../../state/AppState.ts";
@@ -30,7 +31,7 @@ type StaticTranscriptItem =
 			update?: StartupUpdateInfo | null;
 	  };
 
-export function StaticTranscript({
+function StaticTranscriptComponent({
 	items,
 	generation,
 	banner,
@@ -71,3 +72,21 @@ export function StaticTranscript({
 		</Static>
 	);
 }
+
+export function shouldReuseStaticTranscript(
+	previous: Props,
+	next: Props,
+): boolean {
+	if (previous.generation !== next.generation) return false;
+	if (previous.items.length !== next.items.length) return false;
+	// <Static> is append-only. Banner changes are picked up when generation
+	// changes and the transcript is intentionally reprinted. During ordinary
+	// App updates, avoid touching Ink's internal_static node because that
+	// bypasses maxFps and forces an immediate render.
+	return previous.items.every((item, index) => item === next.items[index]);
+}
+
+export const StaticTranscript = memo(
+	StaticTranscriptComponent,
+	shouldReuseStaticTranscript,
+);
--- a/src/ui/renderOptions.ts
+++ b/src/ui/renderOptions.ts
@@ -1,11 +1,39 @@
-export const INTERACTIVE_RENDER_MAX_FPS = 60;
+import type { RenderOptions } from "ink";
 
-export function interactiveRenderOptions(exitOnCtrlC: boolean): {
+const DEFAULT_INTERACTIVE_RENDER_MAX_FPS = 60;
+const MAX_INTERACTIVE_RENDER_MAX_FPS = 120;
+
+interface InteractiveRenderOptionsInput {
 	exitOnCtrlC: boolean;
-	maxFps: number;
-} {
+	env?: Partial<Pick<NodeJS.ProcessEnv, "BACKBOARD_MAX_FPS">>;
+}
+
+type InteractiveRenderOptions = Pick<
+	RenderOptions,
+	"exitOnCtrlC" | "incrementalRendering" | "maxFps"
+>;
+
+export function interactiveRenderOptions({
+	exitOnCtrlC,
+	env,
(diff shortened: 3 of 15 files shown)


src/ui/renderOptions.ts
```diff
@@ -0,0 +1,11 @@
+export const INTERACTIVE_RENDER_MAX_FPS = 60;
```
muhammadbalawal: **Raising the ceiling without making frames cheaper may worsen the symptom.**

Ink writes with `stdout.write(...)` and ignores the return value — there is no `'drain'` gating anywhere in `ink.js`. Every `assistant:delta` triggers a store update, and the streaming message lives *outside* `<Static>`, so each delta costs a full Yoga layout + whole-tree render + terminal write. At 12fps that capped at 12 writes/sec; at 60 it is up to 60.

Over SSH or tmux with a tall transcript, ~10KB frames at 60fps queue roughly 600KB/s into Node's stream buffer. The display falls progressively behind real state — the exact choppiness issue #8 reports, amplified.

**The deeper fix is one line away.** Ink 7.0.5 ships `incrementalRendering`, documented as *"only updates changed lines instead of redrawing the entire output. This can reduce flickering and improve performance for frequently updating UIs"* — a verbatim description of this issue. It defaults to `false` and `grep -rn "incrementalRendering" src` returns nothing. This helper is the natural place for it.

**Also worth reconsidering 60.** Ink's default is 30, i.e. a 34ms window — already below the ~50ms perceptual threshold for typing feedback, at half the redraw cost. The PR argues 84ms → 17ms but never considers 34ms, and there is no profiling or A/B behind the jump.

**And an escape hatch.** Hardcoded means SSH/tmux/slow-terminal users cannot recover the old conservative behavior. The repo already has an env convention (`.env.example`); reading `BACKBOARD_MAX_FPS` with a 60 fallback is ~3 lines and would give the new test something behavioral to assert.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function interactiveRenderOptions(exitOnCtrlC: boolean): {
```
muhammadbalawal: **This return type silently disables TypeScript's excess-property check that the deleted inline literals had.**

Confirmed with a `tsc` probe against this repo:

```ts
render(el, { exitOnCtrlC: false, maxFpsTypo: 60 })  // TS2353 — caught
render(el, opts(false))                              // compiles clean — NOT caught
```

The argument is no longer a fresh object literal, so no excess-property check applies. Concretely: if Ink renames or deprecates `maxFps` in a minor bump (`^7.0.5` floats), `bun run validate` stays green, `render()` silently falls back to Ink's default, and the lag from #8 returns with no type or test failure. The inline literals this PR removes would have caught it.

```suggestion
import type { RenderOptions } from "ink";

export function interactiveRenderOptions(
	exitOnCtrlC: boolean,
): Pick<RenderOptions, "exitOnCtrlC" | "maxFps"> {
```

src/entrypoints/cli.tsx
```diff
@@ -430,10 +431,7 @@ async function main(): Promise<void> {
 			/>
 		</ThemeProvider>
 	);
-	const instance = render(app, {
-		exitOnCtrlC: false,
-		maxFps: 12,
-	});
+	const instance = render(app, interactiveRenderOptions(false));
```
muhammadbalawal: **Boolean trap.** A bare `false` here and `true` at :463 gives the reader no way to tell what it controls — plausible misreadings are "interactive", "isAuthScreen", "debug". The removed `exitOnCtrlC: false` was self-documenting.

This one is load-bearing: the two values are deliberately opposite (main handles Ctrl+C itself via `App`'s `onExit`; the auth screen delegates to Ink). Swapping them in a future edit produces no compile error but either makes Ctrl+C dead in the auth screen or bypasses `controller.cancel({ clearQueue: true })` in the main session.

An options object (`interactiveRenderOptions({ exitOnCtrlC: false })`) or two named constants keeps the label at the call site.

tests/RenderOptions.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from "bun:test";
+import {
+	INTERACTIVE_RENDER_MAX_FPS,
+	interactiveRenderOptions,
+} from "../src/ui/renderOptions.ts";
+
+describe("interactive render options", () => {
+	it("uses a responsive frame ceiling for each interactive root", () => {
+		expect(INTERACTIVE_RENDER_MAX_FPS).toBe(60);
```
muhammadbalawal: **This cannot catch the regression it looks like it guards.**

`grep -rln "entrypoints/cli" tests` returns nothing — no test imports `cli.tsx`. Revert both call sites to inline `{ exitOnCtrlC: false, maxFps: 12 }` and the entire suite still passes, including this file. The actual failure mode (a call site drifting back, or the two sites diverging) is invisible here.

It is also a tautology: `expect(INTERACTIVE_RENDER_MAX_FPS).toBe(60)` restates the line it imports, and 60 is hardcoded three more times below. It can only fail when someone *deliberately* retunes the value — never when behavior regresses. `INTERACTIVE_RENDER_MAX_FPS` is exported solely so this test can read it.

A real seam would assert the value handed to a `render` spy, or assert `cli.tsx` imports the helper.

muhammadbalawal (commented):
Reviewed against Ink 7.0.5's actual source. The diagnosis in issue #8 is right — `maxFps: 12` is genuinely why typing lags — but raising the ceiling alone may make it worse on slow terminals, because it multiplies frames without making any frame cheaper.

Verified locally: Ink's own default is **30** (`ink/build/render.js:16`), `incrementalRendering` exists in 7.0.5 and is set nowhere in `src/`, and `shouldClearTerminalForFrame` (`ink/build/ink.js:83`) triggers a full transcript reprint whenever the dynamic region overflows the viewport — which `App.tsx` reaches routinely during a turn.

Suggested landing: `incrementalRendering: true` + `maxFps: 30` + an env override, rather than 60 alone.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,39 @@
+import type { RenderOptions } from "ink";
+
+const DEFAULT_INTERACTIVE_RENDER_MAX_FPS = 60;
```
muhammadbalawal: Measured this on the branch before/after: with a 400-line transcript and an overflowing live region, 12 FPS gives **104** full-transcript repaints/sec and 60 FPS gives **103**. The ceiling isn't what's limiting us.

Ink's throttle is being bypassed. `reconciler.js:109` takes an `onImmediateRender()` escape hatch whenever `isStaticDirty` is set, and `commitUpdate` (`reconciler.js:223`) sets it every time the `internal_static` host node gets an update — i.e. every time `<Static>` re-renders at all. `StaticTranscript` isn't memoized (`App.tsx:1704`), so that's every App render, i.e. every keystroke.

```
Static usage                       maxFps 12          maxFps 60
────────────────────────────────────────────────────────────────
no <Static>                        12 repaints/s      48 repaints/s
app-style (StaticTranscript)      104/s  2.84 MB/s   103/s  2.82 MB/s
stable props                      103/s  2.80 MB/s    99/s  2.71 MB/s
React.memo'd                       13/s  0.36 MB/s    43/s  1.19 MB/s
```

Memoizing the `<Static>` subtree restores throttling and drops output 8x: 2.84 -> 0.36 MB/s at 12 FPS. It also means this PR's knob would start working. Stable props alone don't help — only not re-rendering does.

Order matters: once the bypass is fixed, 12->60 becomes a real 3.3x increase (0.36 -> 1.19 MB/s). Right now it's harmless only because a worse bug is masking it. I'd fix the memoization first, then re-measure before picking a ceiling.

<sub>Harness: real Ink, fake TTY at 24 rows, 400-line static transcript, 30-line live region, ~125 React commits/sec, 1.5s runs.</sub>

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+	};
+}
+
+function resolveInteractiveRenderMaxFps(value: string | undefined): number {
+	const normalized = value?.trim();
+	if (!normalized || !/^\d+$/.test(normalized)) {
+		return DEFAULT_INTERACTIVE_RENDER_MAX_FPS;
+	}
+	const parsed = Number(normalized);
+	if (parsed < 1 || parsed > MAX_INTERACTIVE_RENDER_MAX_FPS) {
```
muhammadbalawal: Invalid overrides resolve to the **maximum**. Ran the real `process.env` path in a subprocess:

```
unset  -> 60      30 -> 30      0 -> 60
10.5   -> 60     121 -> 60   30fps -> 60      -1 -> 60
```

Every malformed attempt to *lower* the rate returns 60, the highest value — the opposite of what the user asked for, silently. `.env.example:7` advertises "(1-120)", which reads as clamping, and `tests/RenderOptions.test.ts:37-46` locks the current behavior in as intended.

Clamp into `[1,120]` instead, and push a line onto the existing `startupWarnings` array (`cli.tsx`) when the value is garbage.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function interactiveRenderOptions({
+	exitOnCtrlC,
+	env,
+}: InteractiveRenderOptionsInput): InteractiveRenderOptions {
+	const maxFpsOverride =
+		env === undefined ? process.env.BACKBOARD_MAX_FPS : env.BACKBOARD_MAX_FPS;
+	return {
+		exitOnCtrlC,
+		incrementalRendering: true,
```
muhammadbalawal: `incrementalRendering: true` is a second behavior change that isn't in the PR description. Ink documents `@default false` (`render.d.ts:69`).

Measured it in isolation and it changes nothing on the path this PR targets — 2.82 vs 2.80 MB/s — because the clear branch at `ink.js:711` writes `clearTerminal + fullStaticOutput + output` straight to stdout and never touches the line-diffing set up at `ink.js:201`.

So it's carrying the desync risk of incremental line-diffing with no measured benefit here, no test that mounts a component in that mode, and no way to switch it off — `BACKBOARD_MAX_FPS` doesn't disable it. Either drop it or land it separately with its own justification.

tests/RenderOptions.test.ts
```diff
@@ -0,0 +1,62 @@
+import { describe, expect, it } from "bun:test";
+import { readFileSync } from "node:fs";
+import { interactiveRenderOptions } from "../src/ui/renderOptions.ts";
+
+describe("interactive render options", () => {
+	it("uses incremental rendering with a responsive default ceiling", () => {
+		expect(interactiveRenderOptions({ exitOnCtrlC: false, env: {} })).toEqual({
```
muhammadbalawal: All 5 tests pass an explicit `env`, so `renderOptions.ts:21`'s `process.env.BACKBOARD_MAX_FPS` branch only ever runs in production. I confirmed it works today, but nothing in the suite would catch a rename or typo.

One test calling `interactiveRenderOptions({ exitOnCtrlC: false })` with no `env` and a stubbed `process.env` covers it.

Separately, line 54's `expect(cliSource).toContain("interactiveRenderOptions({ exitOnCtrlC: false })")` asserts on raw source text — it breaks when Biome rewraps that line, with no behavior change. And `not.toContain("maxFps: 12")` passes for any other hardcoded ceiling.

src/ui/components/StaticTranscript.tsx
```diff
@@ -71,3 +72,21 @@ export function StaticTranscript({
...
+	previous: Props,
+	next: Props,
+): boolean {
+	if (previous.generation !== next.generation) return false;
+	if (previous.items.length !== next.items.length) return false;
+	// <Static> is append-only. Banner changes are picked up when generation
+	// changes and the transcript is intentionally reprinted. During ordinary
+	// App updates, avoid touching Ink's internal_static node because that
+	// bypasses maxFps and forces an immediate render.
+	return previous.items.every((item, index) => item === next.items[index]);
```
muhammadbalawal: **[blocking] Confirmed regression: the session banner is silently dropped on `/new` after `/resume`.**

Reproduced against real ink 7.0.5 + React 19 with an instrumented harness. `/resume` is the only `setShowBanner(false)` site (App.tsx:1073). On the next `/new`: `agent.newThread()` dispatches `clear`, and the `useSyncExternalStore` flush commits generation+1 with `banner=null` **before** the `.then()` runs. The later `setShowBanner(true)` (App.tsx:1462-1466) lands in a second commit where generation and the (empty) items are unchanged — `items.every(...)` on `[]` is vacuously true — so this comparator returns true and the banner never reaches `<Static>`:

```
compare gen 1->2 len 5->0 ... => RENDER   (banner=null)
compare gen 2->2 len 0->0 banner n->Y => SKIP
NEW banner prints: 0
```

Since `CLEAR_VISIBLE_SCREEN` has just wiped the terminal, the user sits at a blank screen with no session card until their first prompt appends an item (length change finally fails the memo). Pre-PR the unmemoized re-render printed the banner immediately.

The comment above ("Banner changes are picked up when generation changes") is therefore already false for this path — the banner change lands one commit *after* the generation change. Suggested fix at the right depth: fold the banner into the items stream (the component already converts it to a `StaticTranscriptItem`), or at minimum treat a `null -> non-null` banner flip as not-equal — that specific transition is safe for `<Static>` (index 0 after the keyed remount) and is exactly the case being swallowed.

tests/RenderOptions.test.ts
```diff
@@ -0,0 +1,70 @@
+import { describe, expect, it } from "bun:test";
+import {
+	interactiveRenderOptions,
+	resolveInteractiveRenderConfig,
+} from "../src/ui/renderOptions.ts";
+
+describe("interactive render options", () => {
+	it("uses a balanced default frame ceiling", () => {
+		expect(resolveInteractiveRenderConfig(undefined)).toEqual({
```
muhammadbalawal: **[blocking] Confirmed: this test reads the real environment and fails whenever `BACKBOARD_MAX_FPS` is set.**

Passing `undefined` explicitly still triggers the default parameter, so this asserts against the live `process.env.BACKBOARD_MAX_FPS`. Verified by execution:

```
BACKBOARD_MAX_FPS=15 bun test tests/RenderOptions.test.ts
(fail) interactive render options > uses a balanced default frame ceiling
```

Bun auto-loads `.env` from cwd, and this PR's own `.env.example` change invites developers to set exactly this variable — anyone who does breaks `bun run validate`. Pass a sentinel that bypasses the default parameter (e.g. call with an empty string for the default-path case) or stub/delete the env var around this test the way the last test in this file already does.

src/entrypoints/cli.tsx
```diff
@@ -228,10 +232,18 @@ async function main(): Promise<void> {
...
 			: [];
+	const interactiveRenderConfig = resolveInteractiveRenderConfig();
+	const renderWarnings = interactiveRenderConfig.warning
+		? [interactiveRenderConfig.warning]
+		: [];
 	// MCP init warnings (unset env vars, skipped/failed servers) are deliberately
 	// kept out of the startup surface — they cluttered every launch. Server status
 	// is still inspectable via /mcp; only the noisy startup emission is dropped.
-	const startupWarnings = [...hookConfig.warnings, ...permissionWarnings];
+	const startupWarnings = [
```
muhammadbalawal, line 240: **Headless runs now surface a warning about a setting that has no effect there.**

`resolveInteractiveRenderConfig()` runs unconditionally at line 235, before the headless branches (lines 392/404), and the `headless` flag computed at line 220 is never consulted. So `BACKBOARD_MAX_FPS=abc backboard --print "hi"` writes `warning: Invalid BACKBOARD_MAX_FPS ...` to stderr via `runHeadless` (lines 552-555), and in `--format json` the `system:warning` bus emit at lines 247-249 lands in the JSON event stream (`JsonEventStream` subscribes via `onAny` and was attached at line 172) — though neither path ever calls ink `render()`. Pre-PR this variable was not read at all, so scripts that treat stderr output as failure see a new spurious warning. Gate `renderWarnings` on `!headless`.

src/entrypoints/cli.tsx
```diff
@@ -462,10 +477,13 @@ async function runAuthScreen(): Promise<boolean> {
...
 	);
-	const instance = render(app, {
-		exitOnCtrlC: true,
-		maxFps: 12,
-	});
+	const instance = render(
+		app,
+		interactiveRenderOptions({
+			exitOnCtrlC: true,
+			maxFps: resolveInteractiveRenderConfig().maxFps,
```
muhammadbalawal: **The auth screen re-resolves the config and silently discards `.warning`.**

`runAuthScreen` is invoked at line 127 — *before* `main()` builds `startupWarnings` at line 235 — and this call keeps only `.maxFps`. So with `BACKBOARD_MAX_FPS=abc` and no saved credentials, the auth screen renders at the default 30fps with no notice anywhere; if the user exits without logging in, the misconfiguration is never reported. Resolve once (module scope or before the auth loop) and pass the config in — that also removes the duplicate resolution. Related: returning `warnings: string[]` instead of `warning?: string` would match the `loadHookConfig`/MCP convention and let both call sites spread it like `...hookConfig.warnings`.

src/ui/components/StaticTranscript.tsx
```diff
@@ -71,3 +72,21 @@ export function StaticTranscript({
 		</Static>
 	);
 }
+
+export function shouldReuseStaticTranscript(
+	previous: Props,
+	next: Props,
+): boolean {
+	if (previous.generation !== next.generation) return false;
+	if (previous.items.length !== next.items.length) return false;
```
muhammadbalawal: **Armed data-loss hazard: a memo-skipped banner *removal* permanently drops the next transcript item.**

ink 7.0.5's `Static` keeps `index` in state and advances it in a `useLayoutEffect` keyed on `[items.length]` only (`node_modules/ink/build/components/Static.js:11-17`). If a future path flips `showBanner` true->false without a same-commit generation bump, the removal render is memo-skipped while Static's index stays at N+1; the next append renders a banner-less array of the *same* length N+1, the effect doesn't fire, and `slice(N+1)` is `[]` — the appended item never prints. Reproduced empirically: exactly one item silently lost, then output resumes.

No current trigger — the sole `setShowBanner(false)` (App.tsx:1073) batches with `hydrate`'s generation bump — but the `/new` banner bug on line 86 proves same-commit batching is not guaranteed across a microtask boundary, so this is one refactor away. Comparing `banner === null` transitions (both directions) closes it.

src/ui/components/StaticTranscript.tsx
```diff
@@ -71,3 +72,21 @@ export function StaticTranscript({
 		</Static>
 	);
 }
+
+export function shouldReuseStaticTranscript(
```
muhammadbalawal: **Maintenance hazard: the comparator is a hand-maintained prop list with no exhaustiveness link to `Props`.**

Any prop added to `Props` later ships silently dead — it compiles, tests pass, and its updates never render, because this function returns true without consulting it. `banner` is the existing proof, and nothing (type, test, or assert) enforces the "banner changes coincide with generation changes" invariant the comment relies on. The deeper fix that removes the hazard entirely: stabilize the inputs at the source — memoize the `staticItems` slice in `useStableTranscriptLayout` and snapshot the banner per generation — so React.memo's default shallow compare is exact and this custom comparator can be deleted; or fold the banner into the items stream so `Props` collapses to `{items, generation}`.

src/ui/components/StaticTranscript.tsx
muhammadbalawal: **Efficiency: this `every()` walks the full static transcript on every App re-render and can never reference-short-circuit.**

`splitResponsiveTranscript` returns `items.slice(0, splitIndex)` — a fresh array per render (TranscriptLayout.ts:19-22, called without `useMemo` from `useStableTranscriptLayout`) — and `Store.appendAssistantDelta` spreads a new `staticItems` array even when `chunks` is empty (Store.ts:560-564), which is most deltas. So during streaming, a thousands-item transcript gets an O(n) identity scan per store emit just to conclude "reuse". Cheap fixes that compose: reuse `state.render.staticItems` when `chunks.length === 0` in the store, memoize the slice on `[items, splitIndex]`, then add `if (previous.items === next.items) return true;` here so the steady state is O(1).

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,55 @@
...
+const MAX_INTERACTIVE_RENDER_MAX_FPS = 120;
+
+interface InteractiveRenderOptionsInput {
+	exitOnCtrlC: boolean;
+	maxFps: number;
+}
+
+type InteractiveRenderOptions = Pick<RenderOptions, "exitOnCtrlC" | "maxFps">;
+
+export function interactiveRenderOptions({
```
muhammadbalawal: **`interactiveRenderOptions()` is a pure identity function.**

It destructures `{exitOnCtrlC, maxFps}` and returns the same two fields; all the real logic lives in `resolveInteractiveRenderConfig`. Inline literals at both `render()` call sites keep full excess-property checking against ink's `RenderOptions` (the `Pick<...>` adds no rename protection a literal lacks — both error if ink drops `maxFps`), and the only semantic delta is making two optional fields required. The new `tests/StaticTranscript.test.tsx` already bypasses this wrapper because it can't express `stdout`/`stdin`/`patchConsole`/`onRender`. Suggest deleting the function, its two type aliases, and the "preserves each root's Ctrl+C behavior" test (which asserts an object equals itself).

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,55 @@
...
+}: InteractiveRenderOptionsInput): InteractiveRenderOptions {
+	return {
+		exitOnCtrlC,
+		maxFps,
+	};
+}
+
+export interface InteractiveRenderConfig {
+	maxFps: number;
+	warning?: string;
```
muhammadbalawal: **Convention mismatch: singular `warning?: string` vs the codebase's `warnings: string[]`.**

`loadHookConfig` returns `warnings: string[]` (src/core/hooks/types.ts:42-46) and MCP config does the same (src/core/mcp/config.ts:97), which is why cli.tsx can spread `...hookConfig.warnings` directly. The singular field forces the ternary-to-array adapter at cli.tsx:236-238, and the cost is already visible in this PR: the auth-screen call site drops the warning because unwrapping the singular field is extra work the array-spread convention makes free.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function resolveInteractiveRenderConfig(
+	value = process.env.BACKBOARD_MAX_FPS,
+): InteractiveRenderConfig {
+	const normalized = value?.trim();
+	if (!normalized) {
+		return { maxFps: DEFAULT_INTERACTIVE_RENDER_MAX_FPS };
+	}
+	if (!/^-?\d+$/.test(normalized)) {
```
muhammadbalawal, line 21: **This is now the repo's third divergent numeric env-parsing idiom.**

strict-regex + clamp + warn here; `Number()` + floor + silent default in `streamServerErrorRetries` (src/providers/backboard/streamRetry.ts:46-53, accepts "2.7" and "1e2" that this rejects); trim + throw in src/config/env.ts. The repo already has a designated shared home for env parsing — src/utils/envFlags.ts ("Shared parsing for boolean-ish env vars") — so a shared `envInt(value, {min, max, fallback})` there, used by both this and streamRetry, keeps the idioms from drifting further. Not blocking.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,55 @@
...
+		return {
+			maxFps: DEFAULT_INTERACTIVE_RENDER_MAX_FPS,
+			warning: `Invalid BACKBOARD_MAX_FPS "${value}"; using ${DEFAULT_INTERACTIVE_RENDER_MAX_FPS}. Expected an integer from ${MIN_INTERACTIVE_RENDER_MAX_FPS} to ${MAX_INTERACTIVE_RENDER_MAX_FPS}.`,
```
muhammadbalawal: **The raw env value is interpolated unsanitized into a string that reaches the terminal.**

Both warning templates (here and line 51) embed the untrimmed, unbounded `value`. The repo's own `sanitizeForTerminal()` (src/utils/terminalSafe.ts — whose header names exactly this OSC/CSI threat and which is applied in inputSummary, outputPreview, ExecuteTool, AskUserPrompt) is not applied. Verified: in `--print` mode `runHeadless` writes the warning verbatim to stderr, so `BACKBOARD_MAX_FPS=$'\x1b]0;pwned\x07abc'` reaches the terminal unfiltered (interactive ink strips all but SGR). Bun auto-loads `.env` from cwd, so an untrusted cloned repo's `.env` is a modest injection vector. One `sanitizeForTerminal(value)` (plus truncation) fixes it. Low-moderate severity.

src/ui/renderOptions.ts
```diff
@@ -0,0 +1,55 @@
+import type { RenderOptions } from "ink";
+
+const DEFAULT_INTERACTIVE_RENDER_MAX_FPS = 30;
```
muhammadbalawal, line 4: **Note (non-blocking): 30 is exactly ink 7.0.5's own default, and the escape hatch is only discoverable in `.env.example`.**

The default itself is reasonable — it matches `ink.js:177` (`options.maxFps ?? 30`) and the memo fix is what actually makes the throttle effective. But ink fire-and-forgets `stdout.write` with no `drain` backpressure handling, so on a saturated SSH/tmux link the display falls behind with nothing adapting, and `BACKBOARD_MAX_FPS` appears nowhere in README/docs. Worth a README mention of the env var for slow-terminal users.

muhammadbalawal (commented):
Full review of fb4f5c8. The core fix is verified real: memoizing `StaticTranscript` does stop Ink's `isStaticDirty` immediate-render path from bypassing `maxFps` (reproduced against ink 7.0.5), the throttle test genuinely pins that regression, and `bun run validate` passes on this head. But two confirmed defects block the merge — a reproduced user-visible regression (banner dropped on `/new` after `/resume`) and a reproduced environment-dependent test failure — plus two smaller warning-routing bugs that should land in the same revision. Details inline; blocking items marked **[blocking]**.

muhammadbalawal:
lgtm

PR #13, opened 2026-08-20
Add custom sub-agent registry with background execution and time budgets

muhammadbalawal:
Sub-agents shared one hardcoded system prompt and inherited the parent's model and tool set wholesale. This makes them definable per project, lets them run past the turn that spawned them, and gives them wall-clock budgets.

## Defining an agent

`.backboard/agents/researcher.md` — frontmatter configures, body becomes the system prompt:

```markdown
---
description: Deep-dives one question, read-only.
tools: [read, grep, glob]
model: anthropic/claude-opus-5
maxRounds: 30
---

You are a read-only research sub-agent. Never modify files.
Report findings as file paths with a one-line summary each.
```

The parent then calls `Agent(subagent_type: "researcher", prompt: "…")`. Identity comes from the file; the task comes from the parent at call time.

## What's in here

**Agent registry** — discovery from `.backboard/agents/*.md` (project) and `~/.backboard/agents` (user), following the existing MCP/hooks config split. Per-agent `tools`, `disallowedTools`, `model`, `maxRounds`, `timeoutMs`, `background`. `subagent_type` becomes a dynamic enum built from the catalog. Invalid files are skipped with a startup warning, never fatal. `worker` and `rlm` remain built-ins, so behaviour is unchanged with no agent files present.

**Background execution** — `background: true` runs an agent past its spawning turn under a supervisor whose cancellation signal is deliberately unlinked from the turn, so cancelling the foreground no longer kills it. Reports arrive through a new `later` tier on the controller's submit queue, selected at drain time so user input can never be starved. Running agents show in the status bar.

**Time budgets** — `RLMBudget` extracted into a shared `RunBudget` and applied to worker agents. On expiry the agent gets one tool-less turn to report what it established rather than having its work discarded, returning `status: timed_out`.

## Fixes found along the way

- **Concurrent spawns were unbounded** — 20 `Agent` calls in one message meant 20 live sessions. Now capped, with a separate pool for background runs.
- **Deadlock risk** — a nested chain holds one permit per level, so `maxConcurrent < maxDepth` could deadlock against itself. Rejected at construction; a test exercises the tight boundary where permits exactly equal depth.
- **`AgentCatalog` was last-write-wins** — precedence only worked because discovery de-duplicated first. The catalog now keeps the first definition itself.
- **Agent reports rendered shredded** — a raw 4000-char slice that the transcript then truncated mid-line. Now routed through `buildOutputPreview` like every other tool.
- **Headless mode promised reports it would cancel** — `--print` exits after one prompt, so background spawns there now run inline.
- **New threads leaked stale reports** — an agent spawned for a discarded conversation would report into the fresh one.

## Design decisions worth reviewing

- **Background is top-level only.** A nested background spawn would outlive the sub-agent that owns it, leaving no one to receive its report. Deeper spawns run inline regardless of the flag.
- **Nesting is opt-in.** An agent that pins `tools` must list `agent` explicitly to keep spawning.
- **`/undo` does not cover background agents.** They outlive their turn's checkpoint, so journaling edits there would write into an already-finalized checkpoint. Capture is disabled instead; the README says to prefer read-only tools for them.
- **Agent names are matched exactly, not normalized.** Lowercasing `Researcher.md` would collide with `researcher.md`, so the rule is stated in the validation error instead.

## Testing

`bun run validate` clean — 1458 pass, 3 skip, 0 fail. Eight new test files covering discovery and parsing, tool-allowlist policy, the semaphore, worker and RLM timeouts, submit-queue priority, the background supervisor, and an end-to-end flow that drives a real `AgentController` + supervisor + `AgentTool` and asserts the spawning turn returns while the agent is still running, that a second user turn is not blocked, and that the report arrives as its own turn afterwards.

Not covered by tests: live Ink rendering, and a model choosing a background agent on its own. Both were exercised manually.

## Review note

The three areas interleave within `AgentTool.tsx`, `SubAgentRunner.ts`, `cli.tsx` and `Store.ts`, so splitting them into separate commits would have produced commits that don't compile. Kept as one commit with the areas separated in the message.

## Follow-up review fix

Discovery rejected only the shape of a `model:` value, not its parts, so `model: /`, `model: anthropic/`, and `model: /claude-opus-5` parsed into a `ModelRef` with an empty half. The agent then looked valid at startup and failed at request time with an unusable model reference. A qualified reference now requires both parts, `inherit` still means "use the session model", and each part is trimmed so `anthropic / claude-opus-5` no longer yields a provider with trailing whitespace. Malformed values produce the documented skip warning naming the offending value. Covered by four rejection cases and a trimming case in `tests/agentDiscovery.test.ts`.

## Rebase note

Supersedes #6. `main` was rewritten after this branch was cut, leaving the two with no common history, so GitHub refused to reopen that PR. The eight commits are rebased onto current `main`; the only conflicts were `src/prompts/tools/agent.tsx` (main replaced the static prompt with the context-aware `buildAgentPrompt()` — the `subagent_type` wording for the registry is carried over onto it) and the startup-warning list in `src/entrypoints/cli.tsx`, which now carries both `agentCatalog.warnings` and main's `renderWarnings`. The set of files touched is identical before and after the rebase.

### DIFF
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -160,10 +160,7 @@ export class Config {
 		const expertBackendMissing =
 			expert?.enabled === true &&
 			expertModel !== null &&
-			this.auth.backboard === null &&
-			!this.auth.providerKeys.some(
-				(entry) => entry.provider === expertModel.provider,
-			);
+			!this.hasBackendFor(expertModel);
 		this.expertEnabled = (expert?.enabled ?? false) && !expertBackendMissing;
 		this.expertModelRef = expertModel;
 		this.expertModelProfile = expertModel
@@ -480,6 +477,10 @@ export class Config {
 		return this.auth.providerKeys.some((entry) => entry.provider === provider);
 	}
 
+	private hasBackendFor(model: ModelRef): boolean {
+		return this.hasBackboardAuth || this.hasProviderKeyFor(model.provider);
+	}
+
 	get hasBackendForCurrentModel(): boolean {
 		return (
 			this.hasBackboardAuth ||
@@ -498,6 +499,13 @@ export class Config {
 		const next = this.auth.backboard ?? anonymousEnv();
 		this.env.apiKey = next.apiKey;
 		this.env.apiUrl = next.apiUrl;
+		if (
+			this.expertEnabled &&
+			this.expertModelRef &&
+			!this.hasBackendFor(this.expertModelRef)
+		) {
+			this.expertEnabled = false;
+		}
 		return this.auth;
 	}
 
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -375,59 +375,77 @@ export class Config {
 		if ("thinking" in next) this.expertThinking = next.thinking;
 	}
 
+	private saveQueue: Promise<void> = Promise.resolve();
+
+	private enqueueSave(op: () => Promise<void>): Promise<void> {
+		const next = this.saveQueue.then(op);
+		this.saveQueue = next.catch(() => undefined);
+		return next;
+	}
+
 	async saveRuntimeSelection(): Promise<void> {
-		const existing = readBackboardConfig(this.persistedConfigHomeDir);
-		await saveBackboardConfig(
-			{
-				...existing,
-				...(this.flags.model === undefined ? { model: this.currentModel } : {}),
-				...(this.flags.thinking === undefined
-					? { thinking: this.currentThinking }
-					: {}),
-				...(this.flags.memory === undefined
-					? { memory: this.currentMemory }
-					: {}),
-				...(this.flags.memoryProfile === undefined
-					? { memoryProfile: this.currentMemoryProfile }
-					: {}),
-			},
-			this.persistedConfigHomeDir,
-		);
+		await this.enqueueSave(async () => {
+			const existing = readBackboardConfig(this.persistedConfigHomeDir);
+			await saveBackboardConfig(
+				{
+					...existing,
+					...(this.flags.model === undefined
+						? { model: this.currentModel }
+						: {}),
+					...(this.flags.thinking === undefined
+						? { thinking: this.currentThinking }
+						: {}),
+					...(this.flags.memory === undefined
+						? { memory: this.currentMemory }
+						: {}),
+					...(this.flags.memoryProfile === undefined
+						? { memoryProfile: this.currentMemoryProfile }
+						: {}),
+				},
+				this.persistedConfigHomeDir,
+			);
+		});
 	}
 
 	async saveExpertPreference(): Promise<void> {
-		const existing = readBackboardConfig(this.persistedConfigHomeDir);
-		const expert: ExpertConfig = { enabled: this.expertEnabled };
-		if (this.expertModelRef) expert.model = this.expertModelRef;
-		if (this.expertThinking !== undefined) {
-			expert.thinking = this.expertThinking;
-		}
-		await saveBackboardConfig(
-			{ ...existing, expert },
-			this.persistedConfigHomeDir,
-		);
+		await this.enqueueSave(async () => {
+			const existing = readBackboardConfig(this.persistedConfigHomeDir);
+			const expert: ExpertConfig = { enabled: this.expertEnabled };
+			if (this.expertModelRef) expert.model = this.expertModelRef;
+			if (this.expertThinking !== undefined) {
+				expert.thinking = this.expertThinking;
+			}
+			await saveBackboardConfig(
+				{ ...existing, expert },
+				this.persistedConfigHomeDir,
+			);
+		});
 	}
 
 	async saveNotifyPreference(): Promise<void> {
-		const existing = readBackboardConfig(this.persistedConfigHomeDir);
-		await saveBackboardConfig(
-			{
-				...existing,
-				notify: this.notifyEnabled,
-			},
-			this.persistedConfigHomeDir,
-		);
+		await this.enqueueSave(async () => {
+			const existing = readBackboardConfig(this.persistedConfigHomeDir);
+			await saveBackboardConfig(
+				{
+					...existing,
+					notify: this.notifyEnabled,
+				},
+				this.persistedConfigHomeDir,
+			);
+		});
 	}
 
 	async saveVerbosePreference(): Promise<void> {
-		const existing = readBackboardConfig(this.persistedConfigHomeDir);
-		await saveBackboardConfig(
-			{
-				...existing,
-				verbose: this.verboseEnabled,
-			},
-			this.persistedConfigHomeDir,
-		);
+		await this.enqueueSave(async () => {
+			const existing = readBackboardConfig(this.persistedConfigHomeDir);
+			await saveBackboardConfig(
+				{
+					...existing,
+					verbose: this.verboseEnabled,
+				},
+				this.persistedConfigHomeDir,
+			);
+		});
 	}
 
 	get hasBackboardAuth(): boolean {
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -202,8 +202,8 @@ export class Config {
 	}
 
 	/** Runtime gate for tools a sub-agent runs; expert mode never narrows it. */
-	isDelegatedToolEnabled(name: string): boolean {
-		return this.delegatedToolPolicy.isRuntimeAllowed(name);
+	isDelegatedToolEnabled(name: string, model?: ModelRef): boolean {
+		return this.delegatedToolPolicyFor(model).isRuntimeAllowed(name);
 	}
 
 	get toolPolicy(): ToolPolicy {
@@ -211,12 +211,18 @@ export class Config {
 	}
 
 	/**
-	 * The sub-agent's policy: it holds the expert model and does implement, so
-	 * it keeps the implementation tools and takes its allow/deny lists from the
-	 * execution model's profile rather than the planner's.
+	 * The sub-agent's policy: it does implement, so it keeps the implementation
+	 * tools and takes its allow/deny lists from the profile of the model it
+	 * runs on rather than the planner's — the execution model by default, or
+	 * the one a custom agent pins with `model:`. A Moonshot agent spawned from
+	 * an OpenAI session must not inherit the OpenAI profile's
+	 * `apply_patch`-instead-of-`edit` shape.
 	 */
-	get delegatedToolPolicy(): ToolPolicy {
-		return this.buildToolPolicy(false, this.executionModelProfile);
+	delegatedToolPolicyFor(model?: ModelRef): ToolPolicy {
+		return this.buildToolPolicy(
+			false,
+			model ? resolveModelProfile(model) : this.executionModelProfile,
+		);
 	}
 
 	private buildToolPolicy(
--- a/src/config/Config.ts
+++ b/src/config/Config.ts
@@ -6,6 +6,7 @@ import {
 	NO_CREDENTIALS_MESSAGE,
 	resolveAuth,
 } from "./auth.ts";
+import type { ExpertConfig } from "./BackboardConfigTypes.ts";
 import { readBackboardConfig, saveBackboardConfig } from "./backboardConfig.ts";
 import {
 	DEFAULTS,
@@ -79,6 +80,10 @@ export class Config {
 	private verboseEnabled = true;
 	private readonly currentMemoryProfile: MemoryProfile;
 	private currentThinking: ThinkingIntent | null | undefined;
+	private expertEnabled = false;
+	private expertModelRef: ModelRef | null = null;
+	private expertModelProfile: ModelProfile | null = null;
+	private expertThinking: ThinkingIntent | null | undefined;
 	private readonly currentFinalVerificationNudge: boolean;
 	private readonly excludedToolNames: string[];
 	private readonly persistedConfigHomeDir: string | undefined;
@@ -137,6 +142,13 @@ export class Config {
 				: parseThinking(this.flags.thinking);
 		this.currentFinalVerificationNudge =
 			this.flags.finalVerification ?? DEFAULTS.finalVerificationNudge;
+		const expert = persistedConfig.expert;
+		this.expertEnabled = expert?.enabled ?? false;
+		this.expertModelRef = expert?.model ?? null;
+		this.expertModelProfile = expert?.model
+			? resolveModelProfile(expert.model)
+			: null;
+		this.expertThinking = expert?.thinking;
 		this.notifyEnabled = persistedConfig.notify ?? false;
 		this.verboseEnabled = persistedConfig.verbose ?? true;
 		this.excludedToolNames = parseExcludedTools(this.flags.excludedTools).map(
@@ -189,15 +201,37 @@ export class Config {
 		return this.toolPolicy.isRuntimeAllowed(name);
 	}
 
+	/** Runtime gate for tools a sub-agent runs; expert mode never narrows it. */
+	isDelegatedToolEnabled(name: string): boolean {
+		return this.delegatedToolPolicy.isRuntimeAllowed(name);
+	}
+
 	get toolPolicy(): ToolPolicy {
+		return this.buildToolPolicy(this.isExpertModeEnabled);
+	}
+
+	/**
+	 * The sub-agent's policy: it holds the expert model and does implement, so
+	 * it keeps the implementation tools and takes its allow/deny lists from the
+	 * execution model's profile rather than the planner's.
+	 */
+	get delegatedToolPolicy(): ToolPolicy {
+		return this.buildToolPolicy(false, this.executionModelProfile);
+	}
+
+	private buildToolPolicy(
+		expertModeEnabled: boolean,
+		modelProfile: ModelProfile = this.currentModelProfile,
+	): ToolPolicy {
 		return new ToolPolicy({
 			profileTools: this.profile.tools,
-			modelTools: this.currentModelProfile.tools,
+			modelTools: modelProfile.tools,
 			excludedTools: this.excludedToolNames,
-			modelExcludedTools: this.currentModelProfile.excludedTools,
+			modelExcludedTools: modelProfile.excludedTools,
 			computerUseEnabled: this.computerUseEnabled,
 			browserUseEnabled: this.browserUseEnabled,
 			skillDiscoveryEnabled: this.skillDiscoveryEnabled,
+			expertModeEnabled,
 		});
 	}
 
@@ -278,6 +312,51 @@ export class Config {
 		this.currentThinking = thinking;
 	}
 
+	/** On only once a model is picked — expert mode with no model is a no-op. */
+	get isExpertModeEnabled(): boolean {
+		return this.expertEnabled && this.expertModelRef !== null;
+	}
+
+	get expertModel(): ModelRef | null {
+		return this.expertModelRef;
+	}
+
+	/** The model that implements: the expert pick, else the `/model` selection. */
+	get executionModel(): ModelRef {
+		return this.isExpertModeEnabled && this.expertModelRef
+			? this.expertModelRef
+			: this.currentModel;
+	}
+
+	/** The profile of whichever model implements — it decides that model's tools. */
+	get executionModelProfile(): ModelProfile {
+		return this.isExpertModeEnabled && this.expertModelProfile
+			? this.expertModelProfile
+			: this.currentModelProfile;
+	}
+
+	get executionThinking(): ThinkingIntent | null | undefined {
+		return this.isExpertModeEnabled
+			? this.expertThinking
+			: this.currentThinking;
+	}
+
+	/** Omitting `model` keeps the remembered pick, so off/on needs no re-pick. */
+	setExpertMode(next: {
+		enabled: boolean;
+		model?: ModelRef | null;
+		thinking?: ThinkingIntent | null | undefined;
+	}): void {
+		this.expertEnabled = next.enabled;
+		if (next.model !== undefined) {
+			this.expertModelRef = next.model;
+			this.expertModelProfile = next.model
+				? resolveModelProfile(next.model)
+				: null;
+		}
+		if ("thinking" in next) this.expertThinking = next.thinking;
+	}
+
 	async saveRuntimeSelection(): Promise<void> {
 		const existing = readBackboardConfig(this.persistedConfigHomeDir);
 		await saveBackboardConfig(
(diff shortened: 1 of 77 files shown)


src/core/checkpoints/CheckpointStore.ts
```diff
@@ -267,6 +339,7 @@ export class CheckpointStore implements CheckpointRecorder {
 		} else {
 			const content = await readFile(path);
 			const hash = await this.blobs.put(content);
+			if (revokedMidCapture()) return;
```
0vp, line 407: [P1] Make revocation atomic with the final journal flush

The new checks cover revocation during `lstat` and blob reads, but revocation can still happen after this check and `append()` while the final `journal.flush()` is pending. I reproduced that interleaving: the finalized foreground checkpoint retained the pre-image, the background post-image was suppressed, and `/undo` restored the completed background file to its old content. Re-check after the flush before returning, or serialize revocation with active capture commits.

src/core/agent/AgentController.ts
```diff
@@ -185,6 +190,10 @@ export class AgentController {
 		assistantId?: string | null;
 		messages: readonly Message[];
 	}): void {
+		// A just-finished background run may already have queued its report
+		// turn here; cancel it so it cannot land in the resumed conversation.
+		this.cancel({ clearQueue: true });
```
0vp: [P1] Cancel background work before activating replacement storage

This cancellation occurs only after `SessionLifecycle.resume()`/`startNew()` has rotated checkpoint and event-log roots. A finishing background agent can start its report turn in that activation window and execute tools against the replacement session before this line runs. I reproduced the replacement log receiving the stale turn and its side-effect tool executing. Move controller/supervisor cancellation before durable-session activation begins.

src/entrypoints/cli.tsx
```diff
@@ -366,10 +377,19 @@ async function main(): Promise<void> {
 		getTranscriptPath: () =>
 			sessionPathsForRoot(sessionLifecycle.current().sessionRoot).clientLog,
 	});
+	// Deferred: the supervisor is built before the controller that consumes its
+	// reports. Reports queue at "later" so they never preempt user input.
+	backgroundSupervisor.setNotifier((report) => {
+		void controller
```
0vp, line 500: [P1] Disable report submission before shutdown begins

A background run can finish after `onExit()` clears controller work but before `flush()` calls `cancelAll()`. `finish()` removes that run before invoking this notifier, so the later `cancelAll()` cannot stop the newly started report turn, and `controller.dispose()` neither cancels nor awaits it. I reproduced disposal returning while the report turn remained active, allowing work to overlap tool, log, checkpoint, LSP, and MCP shutdown. Disable the notifier and cancel the controller again at the start of `flush()`.

src/core/agents/parse.ts
```diff
@@ -0,0 +1,161 @@
...
+		},
+	};
+}
+
+const INVALID = Symbol("invalid");
+type Invalid = typeof INVALID;
+
+function parseModelRef(value: unknown): ModelRef | undefined | Invalid {
+	const raw = stringValue(value);
+	if (!raw || raw === "inherit") return undefined;
```
0vp: [P2] Reject explicit non-string model values

`stringValue()` returns `null` for values such as `model: 123`, `model: null`, or an object, and this branch treats that exactly like omission or `inherit`. The malformed agent is loaded silently and uses the session model instead of producing the documented startup warning. Only an absent field and the literal `inherit` should return no override; other explicit non-string or empty values should be invalid.

0vp (commented):
Full review of PR #13 found four remaining correctness issues. Each was verified against the rebased head, and the three lifecycle/checkpoint issues were reproduced with focused tests.

src/core/checkpoints/CheckpointStore.ts
0vp: [P1] Revoke pre-images completed before handoff

The post-flush check fixes the reported final-flush interleaving, but a capture that fully returns before handoff still escapes revocation. If `recordPreImage()` completes and the run is handed to the background before Write/Edit/ApplyPatch performs its mutation, `revoke()` only closes the gate. The mutation can still complete while `recordPostImage()` is suppressed, leaving the foreground checkpoint's pre-image to make `/undo` overwrite the background result. I reproduced this on `395ca06`. Track completed capture refs in the gated recorder and append compensating revoke records when the gate closes.

src/core/agent/AgentController.ts
0vp: [P1] Await outgoing turns before rotating session storage

Moving cancellation before activation fixes the original ordering issue, but `beginSessionReplacement()` only signals abort and returns. `/new` then immediately starts activation, so an abort-insensitive or already-committing tool can finish after `CheckpointManager.active` and the event logs have switched to the replacement session. I reproduced an outgoing tool completing against the new root on `395ca06`. Make replacement asynchronous and await `controller.settle()` before calling `activate()`, or reject rotation while active work remains.

src/tools/index.ts
```diff
@@ -66,47 +68,62 @@ export function createDefaultTools(deps?: DefaultToolDeps): Tool[] {
...
-			const registry = new ToolRegistry([...availableTools]);
-			const workerBaseTools = deps.config.toolPolicy.visibleTools(registry);
-			for (const tool of workerBaseTools) {
-				if (!NON_DELEGATABLE_AGENT_TOOLS.has(tool.agentName)) tools.push(tool);
-			}
-			if (holder.tool && deps.config.isToolEnabled(holder.tool.agentName)) {
-				tools.push(holder.tool);
-			}
-			return tools;
+			createRuntimeThinkingResolver(executionView(), deps.client),
```
0vp: [P1] Resolve custom-agent capabilities from the overridden model

When an agent sets `model`, `SubAgentRunner` sends its turns to that override, but this resolver and the delegated candidate policy still use `config.executionModel`. I reproduced a Moonshot agent spawned from an OpenAI session requesting OpenAI thinking metadata and receiving the OpenAI tool profile (`apply_patch`, no `edit`). Cross-provider agents can therefore receive unsupported thinking parameters and the wrong tools. Resolve both thinking and tool policy from `definition.model ?? executionModel`.

src/core/agent/SubAgentRunner.ts
```diff
@@ -84,21 +113,34 @@ export class SubAgentRunner {
...
 			? parentRecorder?.scopedToTurn(params.parentTurnId)
 			: undefined;
+		const revocable = scoped ? revocableRecorder(scoped) : undefined;
+		const checkpoints = revocable?.recorder;
+
+		// A budget bounds how long someone waits. With a handoff, or once the
+		// chain is already backgrounded, nobody is: expiry must not abort.
+		const canHandOff = params.onDeadline !== undefined;
+		const noWaiter = params.parentInBackground === true;
+		const budget = RunBudget.start(
```
0vp, line 131: [P1] Start the wall-clock budget before metadata preflight

This budget is created only after `getThinkingResolver()`/`getThinking()` has completed. If model-metadata lookup stalls, an Agent call with `timeout_ms` remains foreground indefinitely and never times out or hands off. I reproduced a 20 ms run remaining blocked past 80 ms while the resolver was pending. Start the budget before asynchronous preflight and propagate or race its signal through that setup.

src/core/agent/SubAgentRunner.ts
```diff
@@ -115,29 +157,136 @@ export class SubAgentRunner {
...
+		});
+		if (!handle) {
+			// Nowhere to hand it off to, so enforce the budget after all.
+			budget.abortForTimeout();
+			return settle;
+		}
+
+		// The turn no longer owns this run, so its cancellation must not reach
+		// it, and its progress must stop touching a tool row that is now closed.
+		handedOff = true;
```
0vp, line 298: [P1] Mark the live context backgrounded after deadline adoption

Handoff updates `handedOff` but leaves the continuing run's existing `ctx.inBackgroundChain` false. If it invokes Agent afterward, that descendant is passed `parentInBackground: false` and its timeout is enforced, contrary to the documented rule that budgets stop below a run moved to the background. I reproduced a tool executed after handoff still observing the foreground value. Update the live context when adoption succeeds.

src/core/checkpoints/CheckpointStore.ts
```diff
@@ -86,6 +88,68 @@ export interface CheckpointRecorder {
...
+			return noop();
+		},
+		recordPostImage: (absPath, ctx, contentIfInMemory) =>
+			gate.live
+				? recorder.recordPostImage(absPath, gatedCtx(ctx), contentIfInMemory)
+				: noop(),
+		revertToolCall: (...args) => recorder.revertToolCall(...args),
+		beginShellCapture: (cwd, ctx) =>
+			gate.live ? recorder.beginShellCapture(cwd, gatedCtx(ctx)) : noop(),
+		endShellCapture: (ctx) =>
```
0vp: [P2] Release shell snapshots after recorder revocation

If Execute begins shell capture before deadline handoff and finishes afterward, revocation makes this cleanup call a no-op. Its entry remains in `shellBegins`; I reproduced the snapshot still allocated after `endShellCapture()`. Repeated handoffs can retain distinct workspace snapshots of up to 25,000 entries for the session lifetime. Keep the journal suppression, but still remove the matching begin snapshot.

0vp (commented):
Reviewed and reproduced on current head `06826e27`. These four findings are new; the two previously reported checkpoint/session races remain in their existing threads.

src/ui/App.tsx
```diff
@@ -954,8 +966,43 @@ export function App({
 		},
 		[openMcpRegistry],
 	);
+	const applyExpertSelection = useCallback(
+		async (selectedModel: ModelInfo, choice: ThinkingChoice) => {
+			config.setExpertMode({
```
0vp, line 1002: [P1] Defer expert-mode changes until the active turn ends

`/settings` remains available while a turn is running, so this immediately changes the live tool policy. The turn has already snapshotted its advertised schemas, but `ToolScheduler` consults `config.isToolEnabled()` when each call executes. I reproduced enabling Expert while a stream was pending and then having its already-advertised `Write` call rejected as disabled. Block or defer this mutation while `controller.hasActiveWork`, or snapshot the execution policy with the turn.

src/state/AppState.ts
```diff
@@ -86,6 +87,8 @@ export interface AppState {
 	pendingAsk: AskUserRequest | null;
 	model: string;
 	permissionMode: PermissionMode;
+	/** Sub-agents still running past the turn that spawned them. */
+	backgroundAgents: BackgroundRunSnapshot[];
```
0vp, line 91: [P2] Clear background-agent rows when replacing sessions

This state is specific to the current session, but the `clear` and `hydrate` branches in `useAgent` spread the previous state without resetting `backgroundAgents`. Session replacement only signals background cancellation and does not await those runs, so the new or resumed session keeps displaying outgoing agents until they eventually settle; an abort-insensitive run can leave the stale row indefinitely. Reset this array during both clear and hydration.

0vp (commented):
Full re-review completed on `cc76fc02`. The previously posted orchestration findings are fixed; these two issues remain.

src/ui/App.tsx
muhammadbalawal: Fixed in 424ef2b. Both mutation sites (toggleExpertMode and applyExpertSelection) now refuse while controller.hasActiveWork with a notice, matching the existing resume guard, so the tool policy can no longer change under a turn's advertised schemas.

src/state/AppState.ts
muhammadbalawal: Fixed in 424ef2b. The clear and hydrate branches of rootReducer now reset backgroundAgents, with a regression test covering both paths.

src/core/checkpoints/CheckpointStore.ts
```diff
@@ -431,6 +612,9 @@ export class CheckpointStore implements CheckpointRecorder {
 		const before = this.shellBegins.get(callKey);
 		this.shellBegins.delete(callKey);
 		if (!before) return;
+		// Revoked before the walk: the snapshot is released above, and nothing
+		// this command changed may be journaled, so skip the re-scan.
+		if (ctx.mayJournal?.() === false) return;
```
0vp: [P1] Refresh the shell baseline before returning on revocation

This releases the per-call snapshot but leaves `shellIndex` and `shellLastStored` describing the workspace before the background command. If that command writes a file and a foreground Execute begins within the two-second reuse window, even a no-op command journals the background write against the stale snapshot. `/undo` then restores the pre-background content, overwriting the completed background result. Refresh or invalidate the shared baseline before returning while still suppressing journaling.

src/ui/App.tsx
```diff
@@ -1298,6 +1355,42 @@ export function App({
...
+		}
+		if (!config.isExpertModeEnabled) {
+			openModelPicker("expert");
+			return;
+		}
+		config.setExpertMode({ enabled: false });
+		agent.notice(
+			`Expert mode off — ${formatModel(config.model)} implements again.`,
+		);
+		void config.saveExpertPreference().catch((err) => {
```
0vp, line 1383: [P2] Serialize the new expert preference save

This save is launched without awaiting it while the settings panel remains interactive. Disabling Expert and immediately toggling Verbose or Notify starts two saves that each read the entire old config and then rewrite it independently, so whichever write wins erases the other change. I reproduced Expert disabled and Verbose enabled failing to persist together. Queue preference writes or save a merged snapshot of the current in-memory settings.

src/tools/AgentTool.tsx
```diff
@@ -133,12 +159,20 @@ export class AgentTool extends Tool<AgentToolInput, AgentToolOutput> {
 		input: AgentToolInput,
 		ctx: ToolContext,
 	): Promise<ToolResult<AgentToolOutput>> {
-		const subagentType = input.subagent_type ?? "worker";
+		const catalog = this.deps.getCatalog();
+		const requested = input.subagent_type ?? WORKER_AGENT_NAME;
+		const definition = catalog.get(requested);
```
0vp, line 175: [P3] Label custom RLM agents from their catalog definition

The catalog resolves an arbitrary custom name to its actual mode here, but the earlier `tool:start` event classifies the raw `subagent_type` and recognizes only the literal names `worker` and `rlm`. A custom RLM agent such as `analyst` is therefore displayed as `Agent (worker)` throughout execution and changes to RLM only when its result arrives. Make the start event resolve the mode through the catalog as well.

0vp (commented):
Found three remaining correctness and UI issues on the current head. Each was reproduced against this commit, and the full validation suite otherwise passes.

src/core/checkpoints/CheckpointStore.ts
muhammadbalawal: Fixed in 4edc30e. The revoked path no longer skips the end re-scan: it shares the normal walk, so shellLastStored and the index baseline are refreshed to the post-command workspace while the existing post-walk check keeps journaling suppressed. Regression test covers the reuse-window attribution.

src/ui/App.tsx
muhammadbalawal: Fixed in 4edc30e. All four preference save methods in Config now run through a promise queue, so each read-modify-write completes before the next begins; a concurrent expert+verbose save test verifies both changes persist.

src/tools/AgentTool.tsx
muhammadbalawal: Fixed in 4edc30e. Tool gained an agentModeForInput hook that AgentTool overrides to resolve the mode from the catalog; toolStartEvent uses it with the old literal classification as fallback for unknown names.

src/core/tools/ToolPolicy.ts
```diff
@@ -33,6 +47,7 @@ export class ToolPolicy {
 		this.excludedToolNames = combineToolExclusions(
 			snapshot.excludedTools,
 			snapshot.modelExcludedTools ?? [],
+			snapshot.expertModeEnabled ? EXPERT_EXECUTION_TOOLS : [],
```
0vp, line 50: [P1] Preserve an approval path for delegated implementation

This exclusion forces Expert mode to delegate every Write, Edit, ApplyPatch, and Execute call, but delegated runs inherit the default `manual` permission context with `interactive: false`. Their implementation calls therefore resolve to `ask` and are hard-denied because permission prompts are unavailable, while spawning Agent itself is auto-allowed. Under the documented default configuration, Expert mode cannot implement or test changes unless the user already has allow rules or switches permission modes. Route these approvals through the interactive parent or otherwise preserve an executable approval path before applying this exclusion.

src/config/Config.ts
```diff
@@ -149,6 +154,13 @@ export class Config {
 				: parseThinking(this.flags.thinking);
 		this.currentFinalVerificationNudge =
 			this.flags.finalVerification ?? DEFAULTS.finalVerificationNudge;
+		const expert = persistedConfig.expert;
+		this.expertEnabled = expert?.enabled ?? false;
```
0vp: [P2] Disable persisted Expert mode when its backend is unavailable

A BYOK-only user can remove or disable the saved Expert model's provider key and restart with a different valid provider key. Startup accepts that credential but keeps Expert enabled here, removes the parent model's implementation tools, and routes every delegated request to the unavailable Expert provider, where it fails with `No enabled ... API key`. Validate the Expert backend while loading this preference, or leave Expert disabled until a reachable model is selected.

0vp (commented):
Found two remaining Expert-mode correctness issues on the current head. Both were reproduced against this commit; the full validation suite otherwise passes.

src/core/tools/ToolPolicy.ts
muhammadbalawal: Fixed in 8051073. PermissionContext gained an escalation channel: a foreground sub-agent's ask now prompts through the interactive ancestor (serialized with the parent's prompts via a shared lock keyed on the root context), and the channel closes once the chain is backgrounded or handed off. Sub-agent contexts remain interactive:false, so the read-only shortcut and the ask_user tool exclusion are unchanged; headless and background runs still deny.

src/config/Config.ts
muhammadbalawal: Fixed in 8051073. Loading the expert preference now checks that the saved model's provider has an enabled key (or Backboard auth is present); otherwise expert loads disabled, the saved model is kept for re-enabling, and a startup warning says why. Covered by a BYOK-only Config test including the re-enable path.

src/core/permissions/resolveToolPermission.ts
```diff
@@ -114,12 +115,7 @@ export async function resolveToolPermission(
 				input as Parameters<typeof tool.permissionContentIsPaths>[0],
 			),
 		);
-		const answer = await promptForPermission(
-			question,
-			raw,
-			ctx.askUser,
-			ctx.signal,
-		);
+		const answer = await promptForPermission(question, raw, prompt, ctx.signal);
```
0vp: [P2] Revalidate escalation after the prompt resolves

A delegated mutation can obtain the parent prompt while foreground, then cross its deadline handoff while this await is waiting for input. Handoff deliberately keeps the run's signal alive, and after the user answers this path returns `allowed: true` without checking that `pctx.escalate()` has since closed. The parent turn may therefore report the agent as backgrounded before the approved mutation executes, after checkpoint capture has been revoked. Recheck that escalation remains available after the await and deny if the chain has moved to the background.

src/config/Config.ts
```diff
@@ -149,6 +155,28 @@ export class Config {
...
+		const expert = persistedConfig.expert;
+		const expertModel = expert?.model ?? null;
+		const expertBackendMissing =
+			expert?.enabled === true &&
+			expertModel !== null &&
+			this.auth.backboard === null &&
+			!this.auth.providerKeys.some(
+				(entry) => entry.provider === expertModel.provider,
+			);
+		this.expertEnabled = (expert?.enabled ?? false) && !expertBackendMissing;
```
0vp, line 164: [P2] Re-gate Expert mode after live credential changes

This backend gate runs only during construction. If a keys-only session starts with an Anthropic planner and OpenAI Expert model, disabling the OpenAI key through `/keys` calls `refreshAuth()` but leaves Expert enabled, keeps `executionModel` on OpenAI, and continues hiding the parent's implementation tools even though `ByokClient` can no longer resolve that key. The next delegated implementation request fails until restart or manual reconfiguration. Apply the same backend check when authentication refreshes.

0vp (commented):
Found two remaining lifecycle issues in the latest Expert-mode fixes. Both were reproduced against this commit; the full validation suite otherwise passes.

src/core/permissions/resolveToolPermission.ts
muhammadbalawal: Fixed in d832c32. The escalated path re-resolves pctx.escalate() immediately after the prompt await and returns the unavailable denial if the chain has since backgrounded, before acting on the answer — so a stale approval neither executes nor persists an allow rule.

src/config/Config.ts
muhammadbalawal: Fixed in d832c32. The backend check is extracted into hasBackendFor and refreshAuth now applies it, disabling expert in memory when the saved model's provider loses its key; the /keys refresh path in App notices the change. The persisted preference is left intact, so a restart with the key restored re-enables expert.

0vp (commented):
Found two remaining lifecycle/UI issues on the current head. Both were reproduced against this commit; the full validation suite otherwise passes.

src/core/agent/SubAgentRunner.ts
```diff
@@ -84,60 +112,259 @@ export class SubAgentRunner {
...
+			// Nowhere to hand it off to, so enforce the budget after all.
+			budget.abortForTimeout();
+			return settle;
+		}
+
+		// The turn no longer owns this run, so its cancellation must not reach
+		// it, and its progress must stop touching a tool row that is now closed.
+		// Also moves `backgroundChain` to the background, so budgets stop being
+		// enforced for anything the run spawns from here on.
+		handedOff = true;
```
0vp, line 298: [P2] Settle escalated prompts before deadline handoff

If the deadline expires while a delegated mutation is awaiting parent approval, this handoff leaves `budget.signal` live and returns `backgrounded`. The parent turn can then finish, and its cleanup merely clears the AskUser resolver without rejecting the promise or emitting `input:response`. The adopted run remains blocked indefinitely, the stale permission prompt stays visible, and answering it is ignored. Reject or otherwise settle outstanding escalations when ownership moves to the background.

src/ui/hooks/useAgent.ts
```diff
@@ -40,6 +40,7 @@ function rootReducer(state: AppState, action: Action): AppState {
 				transcript: [],
 				todos: [],
 				usage: {},
+				backgroundAgents: [],
```
0vp: [P2] Preserve running background rows during compaction

This `clear` action is also used after `controller.compact()`, which is allowed while background agents are running. It removes their status rows even though compaction does not cancel the supervisor's runs; the eventual `agent:background_finished` event only filters rows and cannot restore them. Manual or automatic compaction can therefore hide active work for the rest of each run. Reset these rows only during session replacement, not transcript-only compaction.

src/core/agent/SubAgentRunner.ts
muhammadbalawal: Fixed in deef059. Each run carries an escalation AbortController combined (AbortSignal.any) into every escalated prompt's signal; the handoff aborts it alongside the other ownership detachments, which rejects the pending ask through the controller's abort path (clearing the UI prompt via input:response). resolveToolPermission catches that rejection and, since the channel is closed, returns the standard unavailable denial instead of throwing — the adopted run continues with a denied call rather than blocking.

src/ui/hooks/useAgent.ts
muhammadbalawal: Fixed in deef059. The clear action is now scoped: newThread and session replacement dispatch scope 'session' (rows reset), the compaction path dispatches scope 'transcript' (rows kept). Regression tests cover both scopes.

0vp (commented):
LGTM

PR #16, opened 2026-08-24
Strip ANSI codes in ContextPanel test

muhammadbalawal:
Update tests/ContextPanel.test.tsx to remove ANSI escape sequences before measuring rendered line lengths. Adds import of stripAnsi from src/utils/terminalSafe.ts and applies it to each line so length checks ignore color/formatting codes, avoiding false failures when asserting narrow-terminal rendering.


0vp:
LGTM

PR #17, opened 2026-08-25
Rebuild computer use with native helpers and evaluation tooling

0vp:
## Summary

- rebuild computer use around a batched runtime, unified tool contract, fresh observations, stable accessibility targeting, and shared browser/platform behavior
- add persistent native macOS and Windows helpers for screenshots, accessibility, keyboard, mouse, application launch, settling, and robust cancellation/lifecycle handling
- attach tool-result images natively across BYOK providers, enforce provider image capabilities, and surface actionable streamed provider errors
- add Daytona and fixture-based CUA evaluations, grounding and smoke scripts, schema verification, metrics, cleanup, and stricter CLI/result validation
- harden session isolation, helper recovery, coordinate handling, image pruning, round limits, and provider catalog behavior with regression coverage

## Validation

- `bun run validate`
- 1,462 tests passed, 9 skipped, 0 failed
- production and script type checks passed
- lint passed
- Swift helper parse passed
- worktree and merge diff checks passed


muhammadbalawal (commented):
Review of the computer-use rebuild — 11 findings, all reproducible from the diff. Split into merge-blockers and narrower bugs; inline comments are tagged accordingly.

**Blockers (5)**
- `ComputerTypes.ts:62` — `move` in both the read-only and state-changing lists → permission gate bypass
- `ComputerPlatformAction.ts:199` — bounds wiped on any miss, unrecoverable for the session
- `ComputerRuntime.ts:203` — `dispose()` reinstalls the platform it just disposed
- `OpenRouterAdapter.ts:149` — `acceptsImages` defaults false → model silently blind
- `OpenRouterAdapter.ts:135` — unguarded catalog `await` turns a `/models` blip into a failed completion

**Bugs (6)**
- `cuaHelper.ps1:272` — missing x/y clicks (0, 0) and reports success
- `ComputerObserver.ts:67` — unguarded `rm()` in a voided promise → unhandled rejection
- `ComputerRuntime.ts:366` — AX re-read with no `settle()` (upstream trigger for the bounds wipe)
- `HelperProcess.ts:93` — abort holds the drain for up to ~65s
- `cuaHelper.ps1:312` — surrogate pairs split across `SendInput` calls
- `MacHelperBinary.ts:61` — EXDEV on cross-filesystem `rename`

Three of these are the same bug fixed on one OS but not the other (surrogate typing, coordinate guards, cross-FS rename) — a parity pass over `cuaHelper.swift` vs `cuaHelper.ps1` may turn up more. The bounds-wipe and the missing `settle()` are one failure chain and are best fixed together.

Four lower-priority cleanups (redundant JSON round-trips in `toolImages.ts`, a docstring claiming helper pruning that doesn't exist, dead exports in `ComputerTool.tsx`, and an `instanceof HelperPlatform` check that belongs on the `Platform` contract) were left off this pass.

src/core/computer/ComputerTypes.ts
```diff
@@ -1,70 +1,130 @@
...
 	| { action: "openApp"; appName: string };
 
-export interface ComputerObservation extends ImageContentPayload<"image/png"> {
-	success: true;
-	action: "screenshot";
+export type ComputerActionName = ComputerAction["action"];
+
+/** Actions that never change machine state. */
+export const READ_ONLY_COMPUTER_ACTIONS: ReadonlySet<ComputerActionName> =
+	new Set<ComputerActionName>(["screenshot", "zoom", "wait", "move"]);
```
muhammadbalawal, line 62: **Blocker — permission gate bypass.**

`move` is listed in both `READ_ONLY_COMPUTER_ACTIONS` (this line) and `STATE_CHANGING_COMPUTER_ACTIONS` (line 69). The two lists disagree about the same action:

- `ComputerTool.isReadOnly` returns `true` for a move-only batch, so `PermissionEngine.ts:74` (`if (!strictManual && tool.isReadOnly(input)) return { behavior: "allow" }`) auto-approves it with no prompt.
- `ComputerRuntime` simultaneously treats it as mutating — marks `elementsDirty` and forces a settle+observe.

A batch of `move`s can open hover menus and continue an in-progress drag, so this silently exempts it from the gate every other input action goes through. Pick one list.

src/core/computer/ComputerPlatformAction.ts
```diff
@@ -1,61 +1,215 @@
...
+					!used.has(candidate.id) &&
+					candidate.role === element.role &&
+					(element.name === undefined ||
+						candidate.name === undefined ||
+						candidate.name === element.name) &&
+					candidate.bounds &&
+					element.bounds &&
+					overlaps(candidate.bounds, element.bounds),
+			);
+		if (!match?.bounds) return { ...element, bounds: undefined };
```
muhammadbalawal, line 199: **Blocker — an element that loses its bounds can never get them back.**

`if (!match?.bounds) return { ...element, bounds: undefined }` clears bounds on *any* unmatched element, and nothing restores them.

Scenario: batch `[click el_3, click el_7]`. Between the two, `observer.accessibility()` samples the AX tree mid-transition and `el_7`'s node is momentarily absent (or it's an unnamed icon, so `named` is empty). Bounds get wiped → `ObservationTargetResolver` throws *"Element el_7 has no screen bounds."*

It's unrecoverable, not transient: with `element.bounds === undefined`, the overlap fallback (`... && element.bounds && overlaps(...)`) can never match again, and the name path only recovers when exactly one candidate shares the name. The element stays unclickable by id until a fresh screenshot resets `state.liveElements`.

Suggest keeping the previous bounds on a miss (`return element`) rather than clearing them.

src/core/computer/ComputerRuntime.ts
```diff
@@ -1,168 +1,455 @@
...
-		if (action.action === "screenshot") {
-			const observation = await new ScreenCapture(
-				ctx.sessionId,
-				this.platform,
-			).capture(ctx.signal);
-			this.lastObservation = observation;
-			return observation;
+	async dispose(): Promise<void> {
+		const platform = this.platformInstance;
+		this.platformInstance = this.options.platform ?? null;
```
muhammadbalawal, line 203: **Blocker — `dispose()` bricks the runtime when a platform was injected.**

```ts
const platform = this.platformInstance;
this.platformInstance = this.options.platform ?? null;
...
await platform?.dispose();
```

With `new ComputerRuntime({ platform })` (cua-smoke.ts, grounding.ts, every test), this disposes the injected platform — setting `HelperPlatform.disposed = true` — and then reinstalls **that same disposed object** as `platformInstance`. Every subsequent `execute()` rejects with *"platform has been disposed"*.

The no-injection path resets to `null` and correctly rebuilds; only the injected path is broken. Separately: the runtime disposes a resource it doesn't own — an injected platform's lifetime belongs to the caller.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -120,17 +130,26 @@ async function* streamOpenRouter(
...
 	void loadCatalog(key)
-		.then((catalog) => {
+		.then((loaded) => {
 			modelMetadata.value =
-				catalog.models.find((candidate) => candidate.name === request.model) ??
+				loaded.models.find((candidate) => candidate.name === request.model) ??
 				null;
 		})
 		.catch(() => {});
+	const acceptsImages = catalog?.imageInputModels.has(request.model) === true;
```
muhammadbalawal, line 149: **Blocker — vision-capable models can silently go blind.**

`imageInputModels` is only populated inside `for (const model of response.data)` behind `if (!isOpenRouterChatModel(model)) continue`. If the configured model is served by OpenRouter but filtered out by that guard (or absent from the catalog page), `catalog?.imageInputModels.has(request.model) === true` is `false`.

`toOpenRouterMessages` is then called with `withImages=false` and every tool screenshot is replaced with `"omitted: model does not accept images"`. Computer use *appears* to work — the model just can't see anything, and no error is surfaced anywhere.

Failing open (assume images are accepted when the model isn't in the catalog) or logging a warning would both be better than silent blindness.

src/providers/byok/adapters/OpenRouterAdapter.ts
```diff
@@ -120,17 +130,26 @@ async function* streamOpenRouter(
 	request: ByokStreamRequest,
 	key: string,
 ): AsyncIterable<ProviderEvent> {
-	const modelMetadata: { value: ModelCatalogItem | null } = { value: null };
+	let catalog = cachedCatalogValue(key);
+	if (!catalog && requestContainsImages(request)) {
+		catalog = await waitForSignal(loadCatalog(key), request.signal);
```
muhammadbalawal, line 135: **Blocker — regression: a catalog blip now kills the whole completion.**

`catalog = await waitForSignal(loadCatalog(key), request.signal)` has no `try`/`catch`. On a cold process (no `cachedCatalogValue`) with a request carrying a screenshot, a 500 or timeout from `/models` rejects `loadCatalog` — the stale-value fallback only applies when a previous value exists — so the generator throws before yielding a single event.

Before this PR the catalog load was fire-and-forget (`.catch(() => {})`) and an outage never affected streaming. Wrapping this in a catch that falls through to the no-catalog path restores that.

src/core/platform/windows/cuaHelper.ps1
```diff
@@ -0,0 +1,392 @@
...
+function Press-Modifiers($mods, $up) {
+  foreach ($m in $mods) {
+    $vk = $script:ModifierKeys[[string]$m.ToLower()]
+    if ($vk) { [Backboard.Native]::Key([uint16]$vk, $up, ($script:ExtendedKeys -contains $vk)) }
+  }
+}
+
+function Invoke-Click($req) {
+  $d = Get-TargetDisplay
+  Move-Cursor $d ([double]$req.x) ([double]$req.y)
```
muhammadbalawal, line 272: **Bug — missing `x`/`y` becomes a click at (0, 0) instead of an error.**

`Move-Cursor $d ([double]$req.x) ([double]$req.y)` — `[double]$null` is `0`, so a request like `{op:"click", button:"left"}` with no coordinates clicks the display's top-left corner (Start button / window system menu) and the helper reports **ok**, so the runtime records success.

`cuaHelper.swift:634` guards the same path: `guard let x = num(params,"x") ... else { throw HelperError("click requires x and y") }`. Worth mirroring here.

Reachability depends on whether `normalizeTargets` can let x/y through undefined — a drag `path` of coordinate arrays that `cleanTarget` passes through unchanged looks like a candidate. If that path is real, this is a blocker rather than a bug; either way the helper shouldn't default to a corner click.

src/core/computer/ComputerObserver.ts
```diff
@@ -0,0 +1,122 @@
...
+		for (const width of FALLBACK_WIDTHS) {
+			if (capture.bytes.byteLength <= MAX_IMAGE_BYTES) break;
+			if (width >= base.maxWidth) continue;
+			capture = await this.platform.screenshot(
+				{ ...base, maxWidth: width },
+				signal,
+			);
+		}
+		this.captures++;
+		if (this.captures % 10 === 0) void this.paths.pruneSession();
```
muhammadbalawal, line 67: **Bug — fire-and-forget prune can take down the process.**

`void this.paths.pruneSession()` / `void this.paths.pruneOldSessions()` — neither has a `.catch()`.

Inside `ComputerPaths.pruneSession`, `readdir` and `stat` are wrapped in try/catch but `await rm(file.path, { force: true })` (line 66) is not. `force: true` swallows ENOENT but **not** EPERM/EBUSY — e.g. a screenshot still open in Preview or Quick Look, or a locked file on Windows. That rejection has no handler, so it surfaces as an unhandled rejection that can terminate the CLI.

A `try`/`catch` around the `rm`, or `.catch(() => {})` on the voided promises, closes it.

src/core/computer/ComputerRuntime.ts
```diff
@@ -1,168 +1,455 @@
...
-			ctx.sessionId,
-			this.platform,
-		).capture(ctx.signal);
-		this.lastObservation = observation;
-		return observation;
+		observer: ComputerObserver,
+		state: ComputerSessionState,
+	): Promise<ObservationTargetResolver> {
+		if (state.lastObservation && state.elementsDirty && usesElementId(action)) {
+			try {
```
muhammadbalawal, line 366: **Bug — AX tree re-read without settling first.**

`resolverFor` calls `await observer.accessibility(ctx.signal)` right after a state-changing action, but `settleScreen()` only runs when the next action is `screenshot`/`zoom` or when `launchedApp` is set.

For `[click Save, click Confirm]` no settle happens, so the AX tree is sampled while the dialog is still animating in. Refreshed bounds are wrong — or missing, which triggers the permanent-wipe bug in `ComputerPlatformAction.ts:199` — and the second click lands off-target.

This PR adds a `settle()` primitive; this is the path that most needs it.

src/core/platform/HelperProcess.ts
```diff
@@ -0,0 +1,240 @@
...
+					this.settle(id);
+					reject(new Error(`${this.label} stdin write failed: ${err.message}`));
+				}
+			});
+		});
+	}
+
+	async dispose(): Promise<void> {
+		this.disposed = true;
+		await this.waitForDrain();
```
muhammadbalawal, line 93: **Bug — abort can block the next request and shutdown for up to ~65s.**

`abortRequest` rejects the caller but keeps the entry pending and chains `this.drain`, which is only released by `settle()` — i.e. when the abandoned helper request finally answers or its timeout fires.

Aborting a `{action:"holdKey", durationMs: 60000}` sets `timeoutMs = durationMs + 5000` in `HelperPlatform.execute`, so the drain can stay held for 65 seconds. During that window the next Computer call blocks in `await this.waitForDrain()` (line 48) and CLI shutdown blocks here in `dispose()`, both with no progress indication.

src/core/platform/windows/cuaHelper.ps1
```diff
@@ -0,0 +1,392 @@
...
+  $d = Get-TargetDisplay
+  if ($null -ne $req.x -and $null -ne $req.y) { Move-Cursor $d ([double]$req.x) ([double]$req.y); Start-Sleep -Milliseconds 30 }
+  $dy = if ($req.dy) { [int]$req.dy } else { 0 }
+  $dx = if ($req.dx) { [int]$req.dx } else { 0 }
+  if ($dy -ne 0) { [Backboard.Native]::Mouse([Backboard.Native]::MOUSEEVENTF_WHEEL, (-$dy * 120)) }
+  if ($dx -ne 0) { [Backboard.Native]::Mouse([Backboard.Native]::MOUSEEVENTF_HWHEEL, ($dx * 120)) }
+}
+
+function Invoke-Type($req) {
+  foreach ($ch in ([string]$req.text).ToCharArray()) {
```
muhammadbalawal, line 312: **Bug — non-BMP characters are corrupted (already fixed on the macOS side).**

`Invoke-Type` iterates `.ToCharArray()` and sends each UTF-16 **code unit** as its own `SendInput` down/up pair with a 4ms gap. For `{action:"type", text:"done ✅🎉"}`, the high and low surrogate of 🎉 are sent as two separate keystrokes.

Windows requires a surrogate pair to arrive as consecutive `KEYEVENTF_UNICODE` inputs with no intervening keyup, so the target app receives two lone surrogates and renders replacement characters.

`cuaHelper.swift:673` has an explicit guard for exactly this (`if end < scalars.count, (0xD800...0xDBFF).contains(...)`); the PowerShell helper has none.

src/core/platform/mac/MacHelperBinary.ts
```diff
@@ -0,0 +1,66 @@
...
+				["-O", "-suppress-warnings", "-o", output, sourcePath],
+				options.signal ?? new AbortController().signal,
+			);
+		} catch (err) {
+			throw new Error(
+				`Could not compile the macOS computer-use helper. Install the Xcode Command Line Tools (xcode-select --install) and retry. ${err instanceof Error ? err.message : String(err)}`,
+			);
+		}
+		await chmod(output, 0o755);
+		await rename(output, target);
```
muhammadbalawal, line 61: **Bug — cross-filesystem `rename` (the same defect fixed on the Windows side).**

`mkdtemp(join(tmpdir(), "backboard-cua-build-"))` builds into the OS temp dir, then `rename(output, target)` moves it under `homedir()`. When `TMPDIR` points at a RAM disk, a separate volume, or a container tmpfs, `rename` throws **EXDEV** and every computer-use action fails with a raw ENOTSUP/EXDEV instead of the helpful Xcode-CLT message.

`WindowsPlatform.ensureWindowsHelperScript` was changed in a later commit to `mkdtemp(join(dir, ".cua-helper-"))` — staging inside the destination directory — precisely to avoid this. Worth mirroring here.

muhammadbalawal:
LGTM

PR #18, opened 2026-08-25
Add Backboard.io docs MCP server

0vp:
## Summary
- add the Backboard.io documentation MCP server to the curated R-CLI catalog
- configure the streamable HTTP endpoint at https://backboard-docs.docsalot.dev/api/mcp

## Validation
- `bun run validate`

### DIFF
--- a/src/core/mcp/catalog.ts
+++ b/src/core/mcp/catalog.ts
@@ -196,6 +196,13 @@ const GENERAL_MCP_SERVERS: readonly McpCatalogServer[] = [
 		description: "Create and manage Supabase projects",
 		url: "https://mcp.supabase.com/mcp",
 	}),
+	remoteServer({
+		name: "backboard-docs",
+		title: "Backboard.io",
+		category: "Project Management & Documentation",
+		description: "Search Backboard documentation",
+		url: "https://backboard-docs.docsalot.dev/api/mcp",
+	}),
 ];
 
 const CLAUDE_REMOTE_MCP_SERVERS: readonly McpCatalogServer[] = [


muhammadbalawal (commented):
Read-only review of the diff. Three things worth addressing before merge: one trust issue on the URL, and two functional bugs that both fall out of the `title` value.

The endpoint itself is fine — verified `initialize` (protocol `2025-06-18`, `serverInfo.name: "backboard-docs"`), `tools/list` returns `list_pages`/`get_page`/`search_docs`, no auth required, and `type: "streamable-http"` maps correctly to `"http"` at `src/core/mcp/config.ts:477`. No name/title/URL collisions against the other 215 entries.

Non-blocking, noted for follow-up: the `find_mcp` confirm prompt renders `item.title` but never `item.detail` (`FindMcpTool.tsx:212`), so the title/host mismatch is invisible at the one point where the user actually consents — pre-existing, worth its own fix. The entry is also appended past its declared category block, and `tests/MCP.test.ts` pins only `slice(0, 21)`, so nothing in the suite covers it.

src/core/mcp/catalog.ts
```diff
@@ -196,6 +196,13 @@ const GENERAL_MCP_SERVERS: readonly McpCatalogServer[] = [
 		description: "Create and manage Supabase projects",
 		url: "https://mcp.supabase.com/mcp",
 	}),
+	remoteServer({
+		name: "backboard-docs",
+		title: "Backboard.io",
+		category: "Project Management & Documentation",
+		description: "Search Backboard documentation",
+		url: "https://backboard-docs.docsalot.dev/api/mcp",
```
muhammadbalawal, line 204: 🚩 **Blocker — first-party branding on a third-party host**

`docsalot.dev` is DocsAlot, an external documentation-platform vendor — not a hostname we control the DNS for. This entry sits in `CURATED_MCP_SERVERS` (the vetted one-click list) and it's the only entry carrying our own brand.

Why this is weightier than a normal catalog addition: MCP tool descriptions land in the system prompt, and `get_page`/`search_docs` output lands directly in the agent's context alongside `Execute`/`Write`/`Edit`. If we churn off DocsAlot, the tenant lapses, or the vendor is compromised, every already-shipped CLI build still offers this as "Backboard.io" — and whoever controls that subdomain at that point gets a prompt-injection channel into user sessions. Configs users have already saved keep pointing there too.

Suggested fix: CNAME a Backboard-controlled hostname at it and ship that instead, e.g. `https://docs.backboard.io/api/mcp`. That decouples the trust from the vendor and makes the URL match the branding.

For the record, the endpoint itself checks out — `initialize` returns protocol `2025-06-18` with `serverInfo.name: "backboard-docs"`, `tools/list` returns `list_pages`/`get_page`/`search_docs`, and `streamable-http` maps correctly to `"http"` at `src/core/mcp/config.ts:477`.

src/core/mcp/catalog.ts
```diff
@@ -196,6 +196,13 @@ const GENERAL_MCP_SERVERS: readonly McpCatalogServer[] = [
 		description: "Create and manage Supabase projects",
 		url: "https://mcp.supabase.com/mcp",
 	}),
+	remoteServer({
+		name: "backboard-docs",
```
muhammadbalawal, line 200: 🐛 **`find_mcp({server: ...})` can never resolve this entry**

`addNamed` (`src/tools/FindMcpTool.tsx:159`) matches only on exact id or exact title:

```ts
s.id.toLowerCase() === target || s.title.toLowerCase() === target
```

With `name: "backboard-docs"` and `title: "Backboard.io"`, there is no string a model or user would plausibly type that hits either one. `find_mcp({server: "backboard"})`, `{server: "Backboard Docs"}`, and `{server: "Backboard Documentation"}` all miss and fall through to:

> No available MCP server named "backboard". Available: &lt;200+ titles&gt;

…which dumps the entire catalog into context.

Every other general entry keeps id ≈ title (`supabase`/`Supabase`, `linear`/`Linear`, `hugging-face`/`Hugging Face`). This is the only one that can't be added by name.

Fix: `title: "Backboard Docs"` — which also fixes the ranking collision on the line below.

src/core/mcp/catalog.ts
```diff
@@ -196,6 +196,13 @@ const GENERAL_MCP_SERVERS: readonly McpCatalogServer[] = [
...
+		title: "Backboard.io",
```
muhammadbalawal, line 201: 🐛 **This title top-ranks for any task mentioning "backboard"**

`tokenize` (`src/tools/discoveryMatch.ts:31`) drops tokens shorter than 3 characters, so `"Backboard.io"` reduces to `{backboard}`. `rankMcpServers` (`src/tools/FindMcpTool.tsx:272`) then scores the title at `strongWeight: 3`.

Net effect: a task like *"fix the backboard auth flow"* scores **3** on this entry and **0** on essentially all 215 others. `best.score > 0`, so `find_mcp` skips straight to `confirmAndAdd` and asks the user to install a documentation-search server for a task that has nothing to do with docs.

This is the only catalog entry whose title collides with the host application's own brand — a string that shows up in a large share of the tasks R-CLI ever sees.

Fix: `title: "Backboard Docs"` (the same change that fixes the `name` mismatch above).

muhammadbalawal:
LGTM

PR #19, opened 2026-08-25
Fix Auto permissions for MCP tools

0vp:
## Summary
- make Auto the default permission mode for new sessions
- keep unannotated and untrusted MCP tools behind the permission gate
- allow only explicitly trusted read-only MCP tools to use the read-only shortcut
- keep local skill discovery gated, while allowing interactive MCP discovery before its own confirmation step
- fail closed to Manual for malformed CLI or project permission modes and surface a project-settings warning
- cycle Shift+Tab from Auto to Accept Edits to Manual
- prevent discovery confirmation prompts in non-interactive execution paths
- update CLI help, README security guidance, and end-to-end regression coverage

## Validation
- `bun run validate`
- 1,577 tests passed, 3 skipped

Stacked on #18.

### DIFF
--- a/src/core/permissions/PermissionMode.ts
+++ b/src/core/permissions/PermissionMode.ts
@@ -31,7 +31,7 @@ export function isKnownPermissionMode(value: string | undefined): boolean {
 }
 
 export function parsePermissionMode(value: string | undefined): PermissionMode {
-	return isKnownPermissionMode(value) ? (value as PermissionMode) : "manual";
+	return isKnownPermissionMode(value) ? (value as PermissionMode) : "auto";
 }
 
 export function nextPermissionMode(mode: PermissionMode): PermissionMode {
--- a/src/core/permissions/PermissionMode.ts
+++ b/src/core/permissions/PermissionMode.ts
@@ -1,7 +1,7 @@
 /**
  * All permission modes. `auto` allows workspace edits, network reads, and any
  * command off the danger list (see dangerousCommands.ts); dangerous commands
- * and outward-facing tools (browser, computer, mutating MCP) still prompt.
+ * and outward-facing tools (browser, computer, destructive MCP) still prompt.
  */
 export const PERMISSION_MODES = [
 	"manual",
--- a/src/core/permissions/PermissionMode.ts
+++ b/src/core/permissions/PermissionMode.ts
@@ -1,7 +1,7 @@
 /**
  * All permission modes. `auto` allows workspace edits, network reads, and any
  * command off the danger list (see dangerousCommands.ts); dangerous commands
- * and outward-facing tools (browser, computer, destructive MCP) still prompt.
+ * and outward-facing tools (browser, computer, mutating MCP) still prompt.
  */
 export const PERMISSION_MODES = [
 	"manual",
@@ -13,7 +13,7 @@ export const PERMISSION_MODES = [
 export type PermissionMode = (typeof PERMISSION_MODES)[number];
 
 /** Modes reachable by Shift+Tab. `bypass` is flag-only (`--permission-mode`). */
-const CYCLE_MODES = ["manual", "acceptEdits", "auto"] as const;
+const CYCLE_MODES = ["auto", "acceptEdits", "manual"] as const;
 
 const LABELS: Record<PermissionMode, string> = {
 	manual: "Manual",
@@ -31,7 +31,8 @@ export function isKnownPermissionMode(value: string | undefined): boolean {
 }
 
 export function parsePermissionMode(value: string | undefined): PermissionMode {
-	return isKnownPermissionMode(value) ? (value as PermissionMode) : "auto";
+	if (value === undefined) return "auto";
+	return isKnownPermissionMode(value) ? (value as PermissionMode) : "manual";
 }
 
 export function nextPermissionMode(mode: PermissionMode): PermissionMode {
--- a/src/tools/FindMcpTool.tsx
+++ b/src/tools/FindMcpTool.tsx
@@ -4,6 +4,10 @@ import type {
 	McpRegistryItem,
 	McpServerRuntimeStatus,
 } from "../core/mcp/index.ts";
+import type {
+	PermissionCheckContext,
+	PermissionDecision,
+} from "../core/permissions/types.ts";
 import { Tool } from "../core/tools/Tool.ts";
 import type { ToolContext } from "../core/tools/ToolContext.ts";
 import { ok, type ToolResult } from "../core/tools/ToolResult.ts";
@@ -77,6 +81,16 @@ export class FindMcpTool extends Tool<Input, Output> {
 
 	override isConcurrencySafe(): boolean {
 		return false;
+	}
+
+	override checkPermissions(
+		_input: Input,
+		ctx: PermissionCheckContext,
+	): PermissionDecision | undefined {
+		if (ctx.mode === "auto") {
+			return { behavior: "allow", reason: "MCP discovery (auto mode)" };
+		}
+		return undefined;
 	}
 
 	override summarizeInput(input: Input): string | undefined {
--- a/src/tools/FindMcpTool.tsx
+++ b/src/tools/FindMcpTool.tsx
@@ -87,7 +87,7 @@ export class FindMcpTool extends Tool<Input, Output> {
 		_input: Input,
 		ctx: PermissionCheckContext,
 	): PermissionDecision | undefined {
-		if (ctx.mode === "auto") {
+		if (ctx.mode === "auto" && ctx.interactive) {
 			return { behavior: "allow", reason: "MCP discovery (auto mode)" };
 		}
 		return undefined;
@@ -210,10 +210,14 @@ export class FindMcpTool extends Tool<Input, Output> {
 	): Promise<ToolResult<Output>> {
 		// Adding a server needs a human confirm; a sub-agent can't prompt, so
 		// surface the candidate instead of throwing on askUser.
-		if ((ctx.agentDepth ?? 0) > 0) {
+		if ((ctx.agentDepth ?? 0) > 0 || ctx.permissions?.interactive === false) {
+			const reason =
+				(ctx.agentDepth ?? 0) > 0
+					? "a sub-agent cannot prompt"
+					: "this run is non-interactive";
 			return ok(
 				{ task: input.task, candidates: [] },
-				`Found the "${item.title}" MCP server, but adding it must be confirmed by the user and a sub-agent cannot prompt. Ask the main agent to run find_mcp for this task.`,
+				`Found the "${item.title}" MCP server, but adding it must be confirmed by the user and ${reason}. Run find_mcp in an interactive main-agent session for this task.`,
 				"Confirmation needed",
 			);
 		}
--- a/src/tools/FindSkillTool.tsx
+++ b/src/tools/FindSkillTool.tsx
@@ -1,8 +1,4 @@
 import { z } from "zod";
-import type {
-	PermissionCheckContext,
-	PermissionDecision,
-} from "../core/permissions/types.ts";
 import type { Skill } from "../core/skills/Skill.ts";
 import type { SkillsShListItem } from "../core/skills/skillsSh.ts";
 import { Tool } from "../core/tools/Tool.ts";
@@ -96,16 +92,6 @@ export class FindSkillTool extends Tool<Input, Output> {
 
 	override isConcurrencySafe(): boolean {
 		return false;
-	}
-
-	override checkPermissions(
-		_input: Input,
-		ctx: PermissionCheckContext,
-	): PermissionDecision | undefined {
-		if (ctx.mode === "auto") {
-			return { behavior: "allow", reason: "skill discovery (auto mode)" };
-		}
-		return undefined;
 	}
 
 	override summarizeInput(input: Input): string | undefined {
@@ -218,10 +204,14 @@ export class FindSkillTool extends Tool<Input, Output> {
 
 		// A download runs third-party code and needs a human confirm; a sub-agent
 		// can't prompt, so surface the candidate instead of throwing on askUser.
-		if ((ctx.agentDepth ?? 0) > 0) {
+		if ((ctx.agentDepth ?? 0) > 0 || ctx.permissions?.interactive === false) {
+			const reason =
+				(ctx.agentDepth ?? 0) > 0
+					? "a sub-agent cannot prompt"
+					: "this run is non-interactive";
 			return ok(
 				{ task: input.task, candidates: remoteCandidates(ranked) },
-				`Found "${c.slug}" on skills.sh, but downloading a third-party skill must be confirmed by the user and a sub-agent cannot prompt. Ask the main agent to run find_skill for this task.`,
+				`Found "${c.slug}" on skills.sh, but downloading a third-party skill must be confirmed by the user and ${reason}. Run find_skill in an interactive main-agent session for this task.`,
 				"Confirmation needed",
 			);
 		}
--- a/src/tools/FindSkillTool.tsx
+++ b/src/tools/FindSkillTool.tsx
@@ -1,4 +1,8 @@
 import { z } from "zod";
+import type {
+	PermissionCheckContext,
+	PermissionDecision,
+} from "../core/permissions/types.ts";
 import type { Skill } from "../core/skills/Skill.ts";
 import type { SkillsShListItem } from "../core/skills/skillsSh.ts";
 import { Tool } from "../core/tools/Tool.ts";
@@ -92,6 +96,16 @@ export class FindSkillTool extends Tool<Input, Output> {
 
 	override isConcurrencySafe(): boolean {
 		return false;
+	}
+
+	override checkPermissions(
+		_input: Input,
+		ctx: PermissionCheckContext,
+	): PermissionDecision | undefined {
+		if (ctx.mode === "auto") {
+			return { behavior: "allow", reason: "skill discovery (auto mode)" };
+		}
+		return undefined;
 	}
 
 	override summarizeInput(input: Input): string | undefined {
--- a/src/tools/MCPToolAdapter.tsx
+++ b/src/tools/MCPToolAdapter.tsx
@@ -1,9 +1,5 @@
 import { z } from "zod";
 import type { McpCallResult, McpToolDefinition } from "../core/mcp/MCPTypes.ts";
-import type {
-	PermissionCheckContext,
-	PermissionDecision,
-} from "../core/permissions/types.ts";
 import type { OpenAITool } from "../core/tools/schema.ts";
 import { Tool } from "../core/tools/Tool.ts";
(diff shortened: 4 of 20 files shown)


src/tools/FindMcpTool.tsx
```diff
@@ -79,6 +83,16 @@ export class FindMcpTool extends Tool<Input, Output> {
 		return false;
 	}
 
+	override checkPermissions(
+		_input: Input,
+		ctx: PermissionCheckContext,
+	): PermissionDecision | undefined {
+		if (ctx.mode === "auto") {
```
muhammadbalawal: **Blocker — headless runs hang forever.**

This allow ignores `ctx.interactive`, and an `allow` verdict short-circuits the guard that protects headless runs. In `resolveToolPermission.ts`, line 76 returns `{ allowed: true }` immediately on `allow`; the `if (!pctx.interactive && !pctx.escalate?.()) return unavailable` check on line 86 only runs on `ask`.

So `execute()` now runs under `--print`. `confirmAndAdd` guards only `agentDepth > 0` (line 213) — the main agent is depth 0 — and falls through to `ctx.askUser`. `AgentController.askQuestions` emits `input:request` and returns a promise settled only by `provideInput()`, but `runHeadless` (`cli.tsx:722`) subscribes to `assistant:delta`, `assistant:message`, `tool:requested`, `turn:end`, `run:error` and `system:warning` — never `input:request`. So `r-cli --print "find an mcp server for postgres"` hangs indefinitely.

Before this PR the tool fell through to `ask` and was denied headless, so `execute()` never ran. With `auto` as the new default this is reachable out of the box. `cli.tsx:320` states the hazard verbatim: *"Both --print and --format json run through runHeadless, which has no UI to answer an input:request. Anything that asks there would hang forever, so the gate must know there is nobody to prompt."*

```suggestion
		if (ctx.mode === "auto" && ctx.interactive) {
```

src/tools/FindSkillTool.tsx
```diff
@@ -94,6 +98,16 @@ export class FindSkillTool extends Tool<Input, Output> {
 		return false;
 	}
 
+	override checkPermissions(
+		_input: Input,
+		ctx: PermissionCheckContext,
+	): PermissionDecision | undefined {
+		if (ctx.mode === "auto") {
```
muhammadbalawal: **Blocker — the same headless hang as `FindMcpTool`.** `tryRemote` reaches `ctx.askUser` (line 229) behind only an `agentDepth > 0` guard, so a `--print` run that calls `find_skill` and falls through to the remote path blocks on a promise nothing resolves. Full mechanism in the comment on `FindMcpTool.tsx`.

```suggestion
		if (ctx.mode === "auto" && ctx.interactive) {
```

**Separately** — the PR description says discovery tools keep "their own confirmation steps", but that only holds for the remote path. `tryLocal` (line 155) and `activateNamed` (line 131) call `activateSkill(...)` and inject the `SKILL.md` body straight into model context with no `askUser` anywhere. Local skills are discovered from `repoSkillRoots(cwd)` — the checked-out repository — so after cloning a hostile repo, in the now-default `auto` mode, `find_skill` silently loads attacker-authored instructions with no user-visible prompt. Previously `isReadOnly() === false` meant the engine fell through to `ask` and the user saw one. The comment on lines 61-63 ("Require a real signal … before silently activating a local skill and injecting its body into context") suggests this path was already understood to be sensitive.

src/tools/MCPToolAdapter.tsx
```diff
@@ -75,6 +79,19 @@ export class McpToolAdapter extends Tool<
 		return this.definition.annotations?.destructiveHint === true;
 	}
 
+	override checkPermissions(
+		input: Record<string, unknown>,
+		ctx: PermissionCheckContext,
+	): PermissionDecision | undefined {
+		if (ctx.mode === "auto" && !this.isDestructive(input)) {
```
muhammadbalawal: **Blocker — a missing `destructiveHint` is treated as "safe", which inverts the MCP spec default and bypasses the trust gate this file applies to `readOnlyHint`.**

`isDestructive()` (line 78) is `annotations?.destructiveHint === true`. Two problems:

1. **The polarity is backwards.** The MCP spec defines `destructiveHint`'s default as **true** for non-read-only tools — absence means destructive, not safe. Most servers ship no annotations at all, so `send_email`, `create_pr`, `delete_record` and friends all return `false` here and auto-run with no prompt in what is now the default mode.
2. **`trustAnnotations` is ignored.** `isReadOnly()` two methods up (line 60) deliberately gates on `this.definition.trustAnnotations`, and there is an existing test — *"ignores readOnlyHint until the user's config trusts the server"* — protecting exactly that. The comment on lines 71-73 spells out the invariant: annotations may be trusted for a scheduling hint, "unlike isReadOnly, which feeds a permission decision." This `checkPermissions` makes `isDestructive` feed a permission decision too, without the gate. The MCP SDK is explicit: *"Clients should never make tool use decisions based on ToolAnnotations received from untrusted servers."* A malicious server needs only to omit the field — the spec default — to get unprompted execution.

Two ways out, either fine:

- **Allowlist (conservative)** — auto-allow only when the user has trusted the server *and* it claims read-only:
  `ctx.mode === "auto" && ctx.interactive && this.definition.trustAnnotations && this.claimsReadOnly()`
- **Keep the denylist, fix the default** — treat absent as destructive and still require trust: `isDestructive` becomes `annotations?.destructiveHint !== false || !this.claimsReadOnly()`, and the auto branch adds `ctx.interactive && this.definition.trustAnnotations`.

Either way `ctx.interactive` is needed here too, for the same headless reason as the discovery tools.

One knock-on worth noting: `checkPermissions` runs above the read-only shortcut in `decidePermission` (`PermissionEngine.ts:67` vs `:74`), so in `auto` mode the `trustToolAnnotations` opt-in stops changing any outcome and `permissionHint` (line 68) can never be shown — there are no prompts left to skip.

src/core/permissions/PermissionMode.ts
```diff
@@ -31,7 +31,7 @@ export function isKnownPermissionMode(value: string | undefined): boolean {
 }
 
 export function parsePermissionMode(value: string | undefined): PermissionMode {
-	return isKnownPermissionMode(value) ? (value as PermissionMode) : "manual";
+	return isKnownPermissionMode(value) ? (value as PermissionMode) : "auto";
 }
```
muhammadbalawal, line 36: **Fails open on malformed input.** Defaulting *absent* input to `auto` is this PR's product decision; defaulting *malformed* input to `auto` is a separate bug. `--permission-mode manul` (or `Manual`, or `readonly`) now silently upgrades privileges — `cli.tsx:331` prints a warning, but the session still runs in `auto`. The settings path is worse: an unrecognized `mode` in `.backboard/settings.json` is dropped, `parsePermissionMode(undefined)` returns `auto`, and **no warning is emitted at all**. A validation fallback should land on the most restrictive mode, not the loosest.

```suggestion
export function parsePermissionMode(value: string | undefined): PermissionMode {
	if (value === undefined) return "auto";
	return isKnownPermissionMode(value) ? (value as PermissionMode) : "manual";
}
```

That keeps `parsePermissionMode(undefined) === "auto"` (the intended new default) while failing closed on typos. `PermissionMode.test.ts:22` would go back to asserting `"manual"` for `"nonsense"`.

muhammadbalawal (commented):
Read-only pass over the diff, then a reproduction run against the branch in a scratch clone.

> **Edited after testing.** My original review listed a headless hang in the discovery tools as a blocker. I tried to reproduce it and could not — it is gated behind a flag that is off in every headless run. Retracted below and in the thread on `FindMcpTool.tsx`. The MCP adapter finding did reproduce.

The direction is reasonable and the pipeline placement is right: putting `checkPermissions` below the deny/ask rules in `decidePermission` preserves the "no mode can skip them" invariant, and `deny and ask rules override MCP auto mode` is the correct test to have added.

## One blocker, reproduced end-to-end

`MCPToolAdapter.tsx:86` treats a missing `destructiveHint` as non-destructive. That inverts the MCP spec default — absent means destructive for non-read-only tools — and bypasses the `trustAnnotations` gate this same file applies to `readOnlyHint`. Driving a real turn with an annotation-less `mailer.send_email` in `auto` mode: the server's `call` ran, no prompt was raised, the turn completed clean. The same call under `acceptEdits` never reaches `call`, so the unprompted execution is new here. Live by default for anyone with an MCP server configured, since this path is not behind the discovery gate. Detail and two suggested fixes in the inline thread.

## Retracted

The headless hang I originally filed against `FindMcpTool.tsx:90` and `FindSkillTool.tsx:105`. The permission gate genuinely is bypassed — `resolveToolPermission` returns `allowed: true` with `interactive: false` — but `ToolPolicy.isRuntimeAllowed` gates both tools behind `skillDiscoveryEnabled`, which defaults to `false` and is only ever set by `toggleDiscovery` in the interactive UI. So they are disabled in every `--print` run, and in an interactive run there is a human to answer. Forcing the flag on with `interactive: false` does hang, so the missing `ctx.interactive` check is worth adding defensively — it is just not blocking.

## Still worth fixing

**`parsePermissionMode` fails open** (`PermissionMode.ts:34`, inline). Verified: `parsePermissionMode("manul")` and `("Manual")` both return `"auto"`. Defaulting absent input to `auto` is this PR's product decision; defaulting malformed input to the loosest mode is a separate bug. The settings path is worse — an unrecognized `mode` in `.backboard/settings.json` is dropped and no warning is emitted at all.

**Shift+Tab now tightens on the first press.** Verified: `nextPermissionMode("auto") === "manual"`. `CYCLE_MODES` is `[manual, acceptEdits, auto]` and the default is now `auto` at index 2, so a user pressing Shift+Tab to grant *more* autonomy — the established meaning of the gesture, and what it did from the previous `manual` default — lands in the strictest mode and has to press twice more to get back. Worth revisiting the cycle order alongside the default flip.

**Deny/ask rules cannot express a server-wide MCP block.** This PR leans on deny/ask as the escape hatch from the new auto-allow, but `RULE_RE` (`PermissionRules.ts:23`) permits no `*` in the tool-name position, `parseRuleSet` filters unparseable rules out silently, and `McpToolAdapter` overrides neither `permissionContent` nor `permissionPaths`. So `{"deny": ["mcp__docs__*"]}` parses to `null` and blocks nothing, with no warning — blocking one server means enumerating every tool it exposes and re-editing settings whenever it adds one. Pre-existing rather than introduced here, but the new test covers only the exact-name case, which reads as assurance that the mitigation works at server granularity.

**Local skill activation has no confirmation** (`FindSkillTool.tsx:131,155`) — narrower than I first wrote it, since it needs the user to have toggled skill discovery on for the session. Details in that thread.

**Docs and coverage.** The rewritten Permissions section never mentions MCP, so this PR's headline behavior change is not user-discoverable; and the security bullet list still names `.backboard/mcp.json` and `.backboard/hooks.json` but not `.backboard/settings.json`, which a cloned repo can use to set `mode: "bypass"`. On tests: every new case hard-codes `interactive: true`, and there is no coverage of a non-interactive session under the new default — which is the configuration where `--print` now runs edits, writes and non-risky commands unattended.

Nit: `PermissionEngine.test.ts:337` hand-rolls an `McpToolDefinition` that `fakeDefinition` (`MCP.test.ts:1688`) already provides.

0vp:
> Read-only pass over the diff — no checkout, nothing run.
> 
> The direction is reasonable and the pipeline placement is right: putting `checkPermissions` below the deny/ask rules in `decidePermission` preserves the "no mode can skip them" invariant, and `deny and ask rules override MCP auto mode` is the correct test to have added.
> 
> Two blockers, both inline:
> 
>     1. **Headless hang** (`FindMcpTool.tsx:90`, `FindSkillTool.tsx:105`) — the auto-mode allow ignores `ctx.interactive`, so `--print` runs reach `ctx.askUser` inside `execute()` and block on a promise `runHeadless` never resolves. One-condition fix.
> 
>     2. **Missing `destructiveHint` read as "safe"** (`MCPToolAdapter.tsx:86`) — inverts the MCP spec default and bypasses the `trustAnnotations` gate the same file applies to `readOnlyHint`, so annotation-less servers (the common case) execute unprompted.
> 
> 
> Plus one more real bug inline on `PermissionMode.ts:34` — invalid mode strings fail open to `auto`.
> 
> Three findings with no good line anchor in this diff:
> 
> **Shift+Tab now tightens on the first press.** `CYCLE_MODES` is `[manual, acceptEdits, auto]` and the default is now `auto` at index 2, so `nextPermissionMode("auto")` wraps to `manual`. A user pressing Shift+Tab to grant _more_ autonomy — the established meaning of the gesture, and what it did from the previous `manual` default — lands in the strictest mode and has to press twice more to get back. Worth revisiting the cycle order alongside the default flip.
> 
> **Deny/ask rules can't express a server-wide MCP block.** This PR leans on deny/ask as the escape hatch from the new auto-allow, but `RULE_RE` (`PermissionRules.ts:23`) permits no `*` in the tool-name position, `parseRuleSet` filters unparseable rules out silently, and `McpToolAdapter` overrides neither `permissionContent` nor `permissionPaths`. So `{"deny": ["mcp__docs__*"]}` parses to `null` and blocks nothing, with no warning — blocking one server means enumerating every tool it exposes and re-editing settings whenever it adds one. Pre-existing rather than introduced here, but the new test covers only the exact-name case, which reads as assurance that the mitigation works at server granularity.
> 
> **Docs and coverage.** The rewritten Permissions section never mentions MCP, so this PR's headline behavior change isn't user-discoverable; and the security bullet list still names `.backboard/mcp.json` and `.backboard/hooks.json` but not `.backboard/settings.json`, which a cloned repo can use to set `mode: "bypass"` — now more reachable given "launch in an unfamiliar repo with no flag" is the common path. On tests: every new case hard-codes `interactive: true`, which is why the headless path above stays green at 1,569 passing.
> 
> Nit: `PermissionEngine.test.ts:337` hand-rolls an `McpToolDefinition` that `fakeDefinition` (`MCP.test.ts:1688`) already provides.

"Read-only pass", actually test and verify please.

src/tools/FindMcpTool.tsx
muhammadbalawal: **Correction — this does not hang. I'm retracting it as a blocker.**

I checked the branch out into a scratch clone and tried to reproduce it. Half the analysis holds: with `mode: "auto"` and `interactive: false`, `resolveToolPermission` really does return `{ allowed: true }`, so the `!pctx.interactive` guard is skipped. But the turn completes anyway, because `find_mcp` never runs headless at all:

```
tool:error  {"name":"FindMcp","error":"Error: tool \"find_mcp\" is disabled"}
turn:end    {"status":"completed"}
```

`ToolPolicy.isRuntimeAllowed` gates `find_mcp` and `find_skill` behind `snapshot.skillDiscoveryEnabled`; `Config.skillDiscoveryEnabled` defaults to `false`; and the only thing that ever sets it is `toggleDiscovery` in `src/ui/App.tsx:1324`, a per-session toggle in the interactive UI that `runHeadless` never constructs. So in a `--print` run these two tools are always disabled, and in an interactive run there is a human to answer the prompt. I could not construct a shipped configuration that reaches the hang.

What's left is latent rather than live: if I force `config.setSkillDiscoveryEnabled(true)` alongside `interactive: false`, the turn hangs exactly as described — `input:request` is emitted and nothing ever resolves it. So adding `ctx.interactive` is still worth doing as a guard against a future caller that enables discovery headlessly, but it is not blocking and shouldn't be treated that way. Apologies for the false alarm.

src/tools/FindSkillTool.tsx
muhammadbalawal: **Correction — same as the `FindMcpTool` thread: no hang, retracting the blocker.** `find_skill` sits behind the same `skillDiscoveryEnabled` gate in `ToolPolicy.isRuntimeAllowed`, which is only ever flipped by `toggleDiscovery` in the interactive UI, so it is disabled in every headless run. Verified by reproduction on the branch. The `ctx.interactive` guard is still worth adding defensively, just not as a blocker.

The second half of the comment stands, with one narrowing I should have caught: `tryLocal` and `activateNamed` do inject a repo-authored `SKILL.md` with no confirmation and, in `auto` mode, no outer prompt either — but reaching it requires the user to have explicitly turned skill discovery on for that session first. So it is a real gap in a path the user opted into, not something a fresh clone hits by default. Lower severity than I first wrote it.

src/tools/MCPToolAdapter.tsx
muhammadbalawal: **This one I did reproduce end-to-end — it holds, and it is the only blocker left.**

Registering an `McpToolAdapter` for a `mailer.send_email` tool with `annotations` absent and `trustAnnotations: false`, then driving a full turn with `mode: "auto"`, `interactive: false`:

- the server's `call` was invoked,
- no `input:request` was ever emitted,
- the turn completed clean, no `tool:error`.

The same call under `mode: "acceptEdits"` never reaches `call` at all, which confirms the unprompted execution is new in this PR rather than pre-existing. And unlike the discovery tools, `McpToolAdapter` is not behind the `skillDiscoveryEnabled` gate that made my other comment a false alarm — this path is live by default for anyone with a server configured.

One correction to my own comment above: the closing line about needing `ctx.interactive` "for the same headless reason as the discovery tools" was wrong. `McpToolAdapter.execute` never calls `askUser`, so there is no hang here. The headless question on this path is a policy one rather than a liveness one — with `auto` as the new default, a `--print` run in CI now executes MCP tools unattended and unprompted. Worth a deliberate call either way, but it is not the same defect.

muhammadbalawal:
## TL;DR

**One real blocker.** `MCPToolAdapter.tsx:86` auto-allows any MCP tool whose server did not mark it `destructiveHint: true`. Most servers ship no annotations at all, so everything they expose auto-runs with no prompt. Two problems: the MCP spec says an absent hint means *destructive*, not safe; and the value is self-reported by the server, which this same file already refuses to trust for `readOnlyHint` (see the comment at line 71). With `auto` as the new default that means `send_email` / `delete_record` / `execute_sql` fire unprompted. Reproduced end-to-end against this branch — the server's `call` ran, no prompt, clean turn; the identical call under `acceptEdits` never reaches `call`, so it is new here. Fix options in the inline thread.

**Also worth fixing, not blocking:**
- `parsePermissionMode` (`PermissionMode.ts:34`) — a typo'd `--permission-mode` falls back to `auto` rather than the safest mode. Verified: `parsePermissionMode("manul") === "auto"`. The settings-file path emits no warning at all.
- Shift+Tab now tightens on the first press. Verified: `nextPermissionMode("auto") === "manual"`, because the new default sits at the end of `CYCLE_MODES`.

**Retracted from my earlier review:** the headless hang I filed against `FindMcpTool` / `FindSkillTool`. `ToolPolicy.isRuntimeAllowed` gates both behind `skillDiscoveryEnabled`, which defaults false and is only set by the interactive UI, so they are disabled in every `--print` run. The missing `ctx.interactive` check is real but unreachable — worth adding defensively, not blocking. Details in those threads.

muhammadbalawal:
LGTM

Issue #21, opened 2026-08-27
Possible fix(deps): 5 vulnerable dependencies in bun.lock


begininvoke:
Came across something in `bun.lock` around line 579 that looked worth flagging.

The project uses shell-quote version 1.8.4, which contains a quadratic‑time parsing bug (CVE‑2026‑13311). An attacker can supply a crafted input string to the parse() function, causing the Node.js event loop to be blocked for a long time (Denial‑of‑Service). No code execution or data leakage occurs, but the availability impact is severe, warranting a HIGH risk rating. Upgrade to a fixed version (>=1.8.5, preferably 1.9.0) to replace the inefficient concat‑based reducer with a linear implementation.

Something like this might fix it:

````diff
```diff
--- a/package.json
+++ b/package.json
@@
-    "shell-quote": "1.8.4",
+    "shell-quote": "^1.9.0",
```
```diff
--- a/bun.lock
+++ b/bun.lock
@@
-    name = "shell-quote"
-    version = "1.8.4"
+    name = "shell-quote"
+    version = "1.9.0"
```
````

For reference: rule `CVE-2026-13311`. Rated high.

The suggested change is untested against this project, so please read it before applying it.

---
*Found with automated scanning ([RedGem](https://code.redgem.net)) and reviewed before opening. If it is not useful, closing it is completely fine.*

PR #22, opened 2026-09-03
Bump shell-quote to 1.10.0

muhammadbalawal:
Closes #21.

Updates the locked `shell-quote` version from 1.8.4 to 1.10.0 to resolve CVE-2026-13311 (quadratic-time `parse()`).

`shell-quote` is not a direct dependency. It is pulled in by `react-devtools-core` (`^1.6.1`) and `@daytona/sdk` (`^1.8.2`), and both ranges already allow the fixed version, so this is a lockfile-only change with no edit to `package.json`.

Verified with `bun install --frozen-lockfile`, `bun run validate`, and `bun audit`, which no longer lists `shell-quote`.


0vp (approved):

0vp:
LGTM

PR #23, opened 2026-09-03
Add configurable custom providers

0vp:
## Summary

- add configurable OpenAI Chat Completions, OpenAI Responses, and Anthropic Messages providers
- add the `/providers` setup and management flow with API-key, environment-variable, and keyless authentication
- support model discovery and manual models, custom headers and request arguments, thinking controls, image inputs, tools, and conversation continuation
- add dynamic BYOK registration, routing, catalog integration, encrypted key storage, and provider lifecycle handling
- harden credential transport, recursive tool schemas, signed tool-call provenance, and Responses continuation metadata isolation
- document custom-provider configuration and extend live schema verification tooling

## Validation

- `bun run validate`
  - 1,636 passed
  - 9 skipped
  - 0 failed
- `bun run build`
- live OpenAI-compatible Chat and Responses probes
- real Ink TUI inference with high thinking
- live Featherless GPT-OSS thinking and Read tool continuation
- live custom Gemini high-thinking and signed Read tool continuation

### DIFF
--- a/scripts/verify-tool-schemas.ts
+++ b/scripts/verify-tool-schemas.ts
@@ -12,6 +12,7 @@
  * Opt-in; never part of `bun test`. Costs a few tokens per model.
  */
 import { Config } from "../src/config/Config.ts";
+import { BYOK_PROVIDER_IDS } from "../src/core/keys/ProviderKeyTypes.ts";
 import { ToolRegistry } from "../src/core/tools/ToolRegistry.ts";
 import { BackboardClient } from "../src/providers/backboard/BackboardClient.ts";
 import type { ModelCatalogItem } from "../src/providers/backboard/types.ts";
@@ -30,12 +31,7 @@ config.enableBrowserUse();
 const router = createAgentClient(config);
 await Promise.all(
 	config.auth.providerKeys
-		.filter(
-			({ provider }) =>
-				!filter ||
-				provider.includes(filter) ||
-				config.modelString.includes(filter),
-		)
+		.filter(({ provider }) => !filter || provider.includes(filter))
 		.map(async ({ provider, key }) => {
 			try {
 				const adapter = config.providerRegistry.get(provider);
@@ -177,8 +173,11 @@ if (backboard) {
 	const server = catalog.filter((m) => m.source !== "byok");
 	const providers = new Set(server.map((m) => m.provider));
 	// The merged catalog hides Backboard's own route for providers you also hold a key for.
-	for (const provider of [...new Set(byok.map((m) => m.provider))])
-		providers.add(provider);
+	for (const provider of BYOK_PROVIDER_IDS) {
+		if (byok.some((model) => model.provider === provider)) {
+			providers.add(provider);
+		}
+	}
 	for (const provider of providers) {
 		if (SKIP.test(provider)) continue;
 		const model =
--- a/scripts/verify-tool-schemas.ts
+++ b/scripts/verify-tool-schemas.ts
@@ -15,7 +15,6 @@ import { Config } from "../src/config/Config.ts";
 import { ToolRegistry } from "../src/core/tools/ToolRegistry.ts";
 import { BackboardClient } from "../src/providers/backboard/BackboardClient.ts";
 import type { ModelCatalogItem } from "../src/providers/backboard/types.ts";
-import { byokAdapter } from "../src/providers/byok/registry.ts";
 import { createAgentClient } from "../src/providers/createAgentClient.ts";
 import { createDefaultTools } from "../src/tools/index.ts";
 
@@ -30,17 +29,26 @@ config.enableComputerUse();
 config.enableBrowserUse();
 const router = createAgentClient(config);
 await Promise.all(
-	config.auth.providerKeys.map(async ({ provider, key }) => {
-		try {
-			await byokAdapter(provider).listModels(key);
-		} catch (err) {
-			throw new Error(
-				`${provider} catalog failed: ${
-					err instanceof Error ? err.message : String(err)
-				}`,
-			);
-		}
-	}),
+	config.auth.providerKeys
+		.filter(
+			({ provider }) =>
+				!filter ||
+				provider.includes(filter) ||
+				config.modelString.includes(filter),
+		)
+		.map(async ({ provider, key }) => {
+			try {
+				const adapter = config.providerRegistry.get(provider);
+				if (!adapter) throw new Error("provider adapter is unavailable");
+				await adapter.listModels(key);
+			} catch (err) {
+				throw new Error(
+					`${provider} catalog failed: ${
+						err instanceof Error ? err.message : String(err)
+					}`,
+				);
+			}
+		}),
 );
 let toolList: ReturnType<typeof createDefaultTools> = [];
 toolList = createDefaultTools({
@@ -71,7 +79,12 @@ const PREFERRED = [
 ];
 
 function pick(models: ModelCatalogItem[]): string | undefined {
-	const names = models.map((m) => m.name).filter((n) => !SKIP.test(n));
+	const names = models
+		.filter(
+			(model) => !filter || `${model.provider}/${model.name}`.includes(filter),
+		)
+		.map((m) => m.name)
+		.filter((n) => !SKIP.test(n));
 	for (const re of PREFERRED) {
 		const hit = names.find((n) => re.test(n));
 		if (hit) return hit;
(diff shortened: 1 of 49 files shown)


src/core/keys/ProviderKeyController.ts
```diff
@@ -86,9 +150,165 @@ export class ProviderKeyController {
...
+			index < 0
+				? [...providers, definition]
+				: providers.map((entry, position) =>
+						position === index ? definition : entry,
+					);
+		await saveBackboardConfig(
+			{
+				...config,
+				providers: next,
+				...((!config.model || !currentModelReachable) && models[0]
```
muhammadbalawal: This branch runs before the `previousId` check, so re-saving a provider that's currently disabled resets `config.model` to `models[0]` even when the selected model is still listed. Repro: disable a provider, edit it, save without changes.

src/config/backboardConfig.ts
```diff
@@ -50,6 +51,7 @@ export function readBackboardConfig(
 			memoryProfile: readMemoryProfileConfig(config),
 			notify: typeof config.notify === "boolean" ? config.notify : undefined,
 			verbose: typeof config.verbose === "boolean" ? config.verbose : undefined,
+			providers: parseCustomProviders(config.providers),
```
muhammadbalawal, line 54: Entries that fail parsing are dropped here, and every writer spreads this result back to disk. A hand-edited provider with a typo gets deleted on the next unrelated save (`/model`, logout, etc.) with no warning.

src/ui/App.tsx
```diff
@@ -308,6 +309,17 @@ export function App({
 	const refreshCredentials = useCallback((): void => {
 		const wasExpert = config.isExpertModeEnabled;
 		refreshClientCredentials(config, client);
+		const persistedModel = readBackboardConfig().model;
```
muhammadbalawal, line 315: This overrides `--model`. In that case the persisted model differs from `config.modelString` on purpose, so toggling any provider in `/providers` switches the session's model.

src/core/keys/ProviderKeyController.ts
```diff
@@ -28,17 +37,65 @@ export class ProviderKeyController {
...
 		const saved = readProviderKeys(this.options.homeDir);
-		return BYOK_ADAPTER_LIST.map((adapter) => {
+		const definitions =
+			readBackboardConfig(this.options.homeDir).providers ?? [];
+		const registry = new ProviderRegistry(definitions, {
+			includeDisabled: true,
+		});
+		const statuses: ProviderKeyStatus[] = registry.adapters.map((adapter) => {
 			const entry = saved[adapter.id];
+			const definition = definitions.find(
```
muhammadbalawal, line 48: A hand-edited config.json entry with a reserved id like `openai` shows up here as a keyless custom provider but never routes (the registry skips reserved ids). Probably better to reject reserved ids in the parser.

src/ui/components/CustomProviderSetup.tsx
```diff
@@ -0,0 +1,577 @@
...
+				}));
+				setStep(3);
+				setChoiceIndex(0);
+				setError(null);
+			} else {
+				const authType = AUTH_OPTIONS[choiceIndex] ?? "apiKey";
+				setDraft((current) => ({
+					...current,
+					authType,
+					credential: "",
```
muhammadbalawal: This wipes the saved variable name when editing an env provider, so it has to be retyped every time.

src/providers/byok/adapters/OpenAIAdapter.ts
```diff
@@ -114,23 +165,41 @@ export function openAIModelAcceptsImages(model: string): boolean {
...
 	// OpenAI caches automatically on the prompt prefix (>1024 tokens), but only
 	// when a request lands on a machine that already holds it. The cache key
 	// pins one conversation to one shard, which is what turns an incidental hit
 	// rate into a reliable one across a long tool loop.
-	if (request.cacheKey) body.prompt_cache_key = request.cacheKey;
+	if (request.cacheKey && options.id === "openai") {
+		body.prompt_cache_key = request.cacheKey;
+	}
+	if (request.maxOutputTokens ?? modelConfig?.maxOutputTokens) {
+		body.max_tokens = request.maxOutputTokens ?? modelConfig?.maxOutputTokens;
```
muhammadbalawal: [P2] Chat adapter sends deprecated `max_tokens`, rejected by OpenAI reasoning models

When a custom `openai-chat` provider configures `maxOutputTokens` on a model, this sends it as `max_tokens`. OpenAI's Chat Completions API (and Azure OpenAI) hard-reject `max_tokens` for reasoning models (o-series, gpt-5 family) with a 400 "Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens'", so every request for such a model fails. Azure OpenAI is exactly the endpoint class this protocol targets. Sending `max_completion_tokens` instead works for reasoning models, and servers that only understand `max_tokens` can still receive it through `extraArgs`. The built-in openai adapter is unaffected only because `request.maxOutputTokens` is never populated and it has no `models` config.

src/config/providers.ts
```diff
@@ -0,0 +1,351 @@
...
+	if (RESERVED_CUSTOM_PROVIDER_IDS.has(id)) return null;
+	const provider: CustomProviderDefinition = {
+		id,
+		name: value.name.trim(),
+		protocol: value.protocol,
+		baseUrl: trimUrl(value.baseUrl),
+	};
+	if (typeof value.enabled === "boolean") provider.enabled = value.enabled;
+	const auth = parseAuth(value.auth);
+	const headers = stringRecord(value.headers);
```
muhammadbalawal, line 106: [P3] One malformed header silently drops all headers — and the loss is persisted

`stringRecord` returns `null` if any single header entry is invalid (non-string value, embedded CR/LF), and `parseProvider` keeps the provider with `headers` omitted entirely — unlike an invalid `baseUrl`, which rejects the whole provider. Consequences: (1) all valid headers, including a required `Authorization`, are silently stripped so requests go out unauthenticated and fail confusingly; (2) `usesCredentials(auth, null)` no longer sees the credential header, so an `auth: none` provider with an `Authorization` header and an http non-localhost baseUrl passes validation it would otherwise fail; (3) because the entry as a whole still parses, the opaque-entry preservation in `saveBackboardConfig` does not protect it, and the next config rewrite persists the entry without its `headers` field, permanently deleting the user's headers. Rejecting the whole provider here (making it opaque, like other invalid fields) fixes all three.

src/providers/byok/registry.ts
```diff
@@ -24,19 +30,94 @@ export const BYOK_ADAPTERS: Record<ByokProviderId, ProviderAdapter> = {
...
+	): string | null {
+		const definition = this.definition(id);
+		if (!definition) return saved?.enabled ? saved.key : null;
+		const auth = definition.auth ?? { type: "apiKey" as const };
+		if (auth.type === "none") return "";
+		if (auth.type === "env") return env[auth.variable]?.trim() || null;
+		return saved?.enabled ? saved.key : null;
+	}
+
+	hasCredential(
```
muhammadbalawal: [P3] Dead exports left behind by the registry migration

`ProviderRegistry.hasCredential` (here), `byokAdapter` (L114), and `byokAdapterFor` (L121) have zero callers across src/, tests/, and scripts/ — every call site now uses `ProviderRegistry`/`BUILTIN_PROVIDER_REGISTRY` directly. `byokAdapter`'s semantics also silently changed from a never-throwing record lookup to one that throws `Unknown provider`, which nothing covers. Remove them, or wire `hasCredential` into the router's `hasKeyFor` check where a credential-presence test would arguably be more correct than `hasProviderKeyFor`.

scripts/verify-tool-schemas.ts
```diff
@@ -30,17 +29,26 @@ config.enableComputerUse();
...
-				}`,
-			);
-		}
-	}),
+	config.auth.providerKeys
+		.filter(
+			({ provider }) =>
+				!filter ||
+				provider.includes(filter) ||
+				config.modelString.includes(filter),
```
muhammadbalawal: [P3] `config.modelString.includes(filter)` makes the `-f` filter a no-op

This clause doesn't depend on the loop item — `config.modelString` is constant for the whole run. Whenever the `-f` value happens to match the configured default model string (e.g. `-f claude` while the default model is `anthropic/claude-...`), the OR short-circuits to true for every entry, so all provider catalogs are fetched instead of only the filtered one, and a catalog failure in an unrelated, filtered-out provider aborts the whole run. Dropping this clause (leaving `!filter || provider.includes(filter)`) restores the documented behavior; the later probes already handle model-name filters via `${provider}/${model}`.

muhammadbalawal (commented):
Follow-up review of d99dd30: the five earlier findings are fixed (branch order in saveDefinition, opaque provider preservation, reserved-id rejection, --model guard, env credential prefill) — nice work. Four smaller issues remain, posted inline below.

muhammadbalawal:
LGTM

muhammadbalawal (approved):

